PR #27, opened 2020-11-04
Added build pipeline to ensure that the build isn't broken

TheKnarf:
Also added a build command and a file to lintignore

### DIFF
--- a/.lintignore
+++ b/.lintignore
@@ -1,1 +1,2 @@
 .cache
+public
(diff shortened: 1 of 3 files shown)


.lintignore
```diff
@@ -1 +1,2 @@
 .cache
+public
```
TheKnarf, line 2: `public` is the folder where gatsby output it's build output

Trulsaa (approved):

PR #29, opened 2020-11-05
Feature/storybook

TheKnarf:
This PR three things:

- First I remove yarn PnP support, I was having problems making PnP work with both gatsby, sanity, and storybook. So until more packages get updated to work properly with PnP I have disabled it. That does mean that after merging this in you will again have a `node_modules` folder in your project.

- Initial setup of Storybook! Here I also created some example components and added a sample story for each of them. That way we got something to work with.

~Lastly, I tried setting up a Github Action for deploying the resulting Storybook to Github pages. We'll see if this works. If this is to much to review in one PR then I could create a PR on its own with just the Github Action related stuff.~ *Moved this  to it's own branch*


.github/workflows/linting.yml
```diff
@@ -1,6 +1,10 @@
 name: Linting
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      -master
+  pull_request:
```
TheKnarf, line 7: This should perhaps be its own comitt / PR.

.github/workflows/deploy-storybook.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: '14.x'
+
+      - name: Install dependencies
+        run: yarn --immutable
+
+      - name: Install dependencies
```
Trulsaa, line 23: Copy paste

Trulsaa (approved):

PR #30, opened 2020-11-05
tailwind config updated with color profiles (issue)

mikkelcodes:
Hei, tenkte det var lettere å gjøre det på denne måten i stedet.


TheKnarf:
This looks great! We just need to get the build working and it's all good. It seems that you forgot to commit the `yarn.lock` file. If you add it then the build should go through :)

TheKnarf (approved):

TheKnarf (approved):

PR #32, opened 2020-11-06
Setup Sanity based on their blog starter.

TheKnarf:


packages/cms/config/.checksums
```diff
@@ -0,0 +1,7 @@
+{
+  "#": "Used by Sanity to keep track of configuration file checksums, do not delete or modify!",
```
TheKnarf, line 2: Am I supposed to have committed this to Git? Or was it supposed to be gitignored?

packages/cms/static/logo.svg
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 23.0.6, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
```
TheKnarf, line 2: Perhaps I should have optimized this before adding it.

mikkelcodes (dismissed):

mikkelcodes (approved):

PR #40, opened 2020-11-06
gatsby/sanity connection and font prefetch feature/#23

mikkelcodes:
Font prefetching og sanity config i gatsby. Det er nå mulig å query data fra sanity-prosjektet (localhost:8000/__graphql)

### DIFF
--- a/packages/website/gatsby-config.js
+++ b/packages/website/gatsby-config.js
@@ -1,3 +1,19 @@
+require('dotenv').config({
+  path: `.env.${process.env.NODE_ENV}`,
+});
+
 module.exports = {
-  plugins: ['gatsby-plugin-postcss'],
+  plugins: [
+    'gatsby-plugin-postcss',
+    {
+      resolve: `gatsby-source-sanity`,
+      options: {
+        projectId: `mnr37rl0`,
+        dataset: `production`,
+        token: process.env.SANITY_TOKEN,
+        graphqlTag: 'default',
+        overlayDrafts: true,
+      },
+    },
+  ],
 };
--- a/packages/website/gatsby-config.js
+++ b/packages/website/gatsby-config.js
@@ -15,5 +15,17 @@ module.exports = {
         overlayDrafts: true,
       },
     },
+    {
+      resolve: `gatsby-plugin-prefetch-google-fonts`,
+      options: {
+        fonts: [
+          {
+            family: `Source Sans Pro`,
+            subsets: [`latin`],
+            variants: [`300`, `400`, `700`],
+          },
+        ],
+      },
+    },
   ],
 };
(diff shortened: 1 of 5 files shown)


packages/website/gatsby-config.js
```diff
@@ -1,3 +1,30 @@
+require('dotenv').config({
+  path: `.env.${process.env.NODE_ENV}`,
```
TheKnarf, line 2: What does these two lines do?

packages/website/package.json
```diff
@@ -7,11 +7,15 @@
...
+    "add": "^2.0.6",
     "gatsby": "^2.25.2",
     "gatsby-plugin-postcss": "^3.0.4",
+    "gatsby-plugin-prefetch-google-fonts": "^1.4.3",
+    "gatsby-source-sanity": "^6.0.4",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
-    "tailwindcss": "^1.9.6"
+    "tailwindcss": "^1.9.6",
+    "yarn": "^1.22.10"
```
TheKnarf, line 18: How did yarn become a dependecy here?

packages/website/package.json
```diff
@@ -7,11 +7,15 @@
     "build:storybook": "build-storybook --no-dll --quiet"
   },
   "dependencies": {
+    "add": "^2.0.6",
```
TheKnarf, line 10: Where do we use this?

TheKnarf (changes requested):
Perhaps we could split this into two commits? One for the `font` issue and one for the `sanity plugin`?

packages/website/package.json
TheKnarf: I think perhaps you accidentally ran `yarn add yarn add` and thereby added `yarn` and `add` as a dependency.
You should be able to fix it by removing those two dependencies, re-run `yarn` and then add and commit it with amend :)

packages/website/gatsby-config.js
mikkelcodes: It allows gatsby-config to access the following file packages/website/.env.development which contains our environment variables (Sanity access token).

.env.development, which should be ignored by .gitignore, should contain one line:
SANITY_TOKEN="token goes here"

Sanity picks up this token on line 11 in gatsby-config.js and this allows for draft previews. I can demo it later when I have some content to preview. I also forgot to add "overlayDrafts: true" to the config, so I will include that in the upcoming ----amend 👍

packages/website/package.json
mikkelcodes: Yup, that's probably what happened here. Got it

mikkelcodes:
> Perhaps we could split this into two commits? One for the `font` issue and one for the `sanity plugin`?
Didn't see your comment here before I pushed the amended commit. If you agree with all changes, could we let it slide this time and I'll keep all future PRs specific to each single issue?

TheKnarf (approved):

PR #41, opened 2020-11-08
title component added

mikkelcodes:
Added some commonly used font-size and line-height combinations to tailwind also


packages/website/src/components/title/Title.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+const Dot = () => <span className="text-yellow">.</span>;
+
+export const Title = ({ title }) => (
+  <h2 className="text-white w-full text-center text-4xl">
+    {title.toUpperCase()}
```
TheKnarf, line 7: Maybe this could be solved with css instead using `text-transform: uppercase;`?

packages/website/src/components/title/Title.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
```
TheKnarf, line 1: I would prefer to keep the filename of this component as `index.js` as it originally was. I also prefer to keep filenames all lowercase.

TheKnarf (changes requested):
Three small comments, otherwise it looks good!

packages/website/src/components/title/Title.js
```diff
@@ -0,0 +1,10 @@
...
+
+const Dot = () => <span className="text-yellow">.</span>;
+
+export const Title = ({ title }) => (
```
TheKnarf, line 5: Instead of `({ title }) =>` it would  probably be more appropriate to use `({ children }) =>` here.

packages/website/src/components/title/Title.js
mikkelcodes: It sure can be solved that way.

packages/website/src/components/title/Title.js
mikkelcodes: Okay, got it. I'll stick to that from now on and update the files.

packages/website/src/pages/index.js
```diff
@@ -1,28 +1,16 @@
 import React from 'react';
-import Layout from '../components/layout/Layout';
+import Layout from '../components/layout/index';
```
TheKnarf, line 2: Since the file is called `index` you can instead import it as just `../components/layout`, you don't need the `index`-part ( that is kinda the point of calling the file index inside a folder ;-) )

packages/website/src/components/title/Title.js
mikkelcodes: If you don't need the possibility of editing page titles in Sanity, it makes sense to update this. Otherwise, it would just be title={data.sanityPage.title} once cms side is configured. This is what I had in mind when I wrote this. Thoughts?

TheKnarf: Would that not work just as well with `<Title>{data.sanityPage.title}</Title>`?

mikkelcodes: Yes it would, its only slightly longer and I thought that <Title title={title}/> was descriptive enough. I will update it once again

TheKnarf: `<Title>{data.sanityPage.title}</Title>` and `<Title title={data.sanityPage.title}/` is exactly the same length ;-)

packages/website/src/pages/index.js
mikkelcodes: Ahh, now it makes total sense with this naming style when components have their own folders. Ill update this

packages/website/src/components/title/index.stories.js
```diff
@@ -0,0 +1,21 @@
...
+      </div>
+    ),
+  ],
+  argTypes: {
+    children: 'text',
+  },
+};
+
+export const Default = ({ children }) => {
+  console.log(children);
```
TheKnarf, line 19: This console.log can be deleted

TheKnarf (approved):
Looking good!

PR #45, opened 2020-11-09
Docker setup

TheKnarf:
Would love your feedback here @Trulsaa since you have worked a bit with Docker.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+node_modules/
+public/
+dist/
--- a/Readme.md
+++ b/Readme.md
@@ -15,6 +15,36 @@ $ yarn --version
  2.3.3
 ```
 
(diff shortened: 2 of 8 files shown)


mikkelcodes:
Could you briefly explain the need for using docker? I will approve it regardless

mikkelcodes (approved):

TheKnarf:
> Could you briefly explain the need for using docker? I will approve it regardless

The idea of using docker is so that I can test that everything works locally before pushing it up to Azure, in theory then everything should be the same locally as on the server. It is a bit overkill to use Docker given that it's all just static files, but it makes the stack more similar to other projects in the [Alv org](github.com/Alv.no).

Trulsaa (approved):

PR #46, opened 2020-11-09
Component/page description

mikkelcodes:


TheKnarf (commented):
You could try squashing those four commits together into one by doing:

`git rebase -i HEAD~4` and then changing them to squash.

PR #54, opened 2020-11-10
add tags component, wip

mikkelcodes:


TheKnarf:
Seems like you forgot to add the `tags` folder with the component?
You may also want to update the branch quickly by running `git fetch origin && git rebase origin/master`.

packages/website/src/svg/TagIcon.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
```
TheKnarf, line 1: Maybe move this into `src/components/icon` instead and call the component `Tag`?

TheKnarf (approved):

PR #55, opened 2020-11-11
add employees backend, data sourcing, filtering feature

mikkelcodes:
I recommend we use gatsby-image for standalone images. This gives us lazy-loading, an optional image blur-up effect (pretty cool imo) and great file-size compression. Sanity has their own gatsby graphql fragments we can use too. 

This is why I have thrown in a few dependencies related to images. I might use this with the employee photos.

Otherwise, the filtering should work fine now


mikkelcodes:
Også verdt å nevne: teksten på bildene er lite synlige i default state. Dette er på todo

packages/cms/schemas/tag.js
```diff
@@ -0,0 +1,13 @@
+export default {
+  name: 'tag',
```
TheKnarf, line 2: Should this have been called `employee_tags` or something similar?

packages/website/package.json
```diff
@@ -9,9 +9,14 @@
...
   "dependencies": {
     "gatsby": "^2.25.2",
+    "gatsby-background-image-es5": "^1.1.2",
+    "gatsby-image": "^2.4.21",
     "gatsby-plugin-postcss": "^3.0.4",
     "gatsby-plugin-prefetch-google-fonts": "^1.4.3",
+    "gatsby-plugin-sharp": "^2.7.1",
     "gatsby-source-sanity": "^6.0.4",
+    "gatsby-transformer-sharp": "^2.5.21",
+    "intersection-observer": "^0.11.0",
```
TheKnarf, line 19: All of these gatsby plugins should probably be moved under `devDependencies`

packages/website/src/components/tagbar/index.js
```diff
@@ -1,84 +1,116 @@
...
       </div>
-      <div className="flex justify-center flex-wrap">
-        {employees.map((employee) => (
-          <div className="">
-            <div>{employee.xp}</div>
-            <div>
-              <div>{employee.name}</div>
-              <div>{employee.title}</div>
+
+      {/* EMPLOYEE CARDS */}
```
TheKnarf, line 78: Can `EmployeeCard` get its own component?

packages/website/src/components/tagbar/index.js
```diff
@@ -1,84 +1,116 @@
...
+      <div className="flex justify-center flex-wrap max-w-1320 mx-auto">
+        {employees.map((employee) => {
+          const details = employee.node;
+          return (
+            <div
+              style={{
+                backgroundImage: `url(${details.image.asset.fluid.src})`,
+                backgroundSize: 'cover',
+                backgroundOpacity: '0',
```
TheKnarf, line 87: If background size and opacity is static, should they not be styled by className?

packages/website/src/components/tagbar/index.js
```diff
@@ -1,84 +1,116 @@
...
+              }}
+              className="p-2 flex cursor-pointer transition opacity-75 hover:opacity-100 filter-grayscale hover:filter-grayscale-0 hover:opacity-100 duration-100 m-2 w-277 h-90"
+              key={details.id}
+            >
+              <div className="absolute -ml-2">
+                <Shade />
+                <div className="z-10 relative transform -translate-y-9 translate-x-2 text-sm flex items-center">
+                  <span className="text-xl font-bold mr-1 pointer-events-none">
+                    {details.experience}
+                  </span>{' '}
```
TheKnarf, line 97: An extra space added here `{' '}` that can be removed.

TheKnarf (changes requested):
A couple of comments, but otherwise this looks good! Mainly I think we perhaps should refactor `tagbar` a bit as it's starting to get a bit big. At least an employee card should have its own component.

packages/website/src/components/tagbar/index.js
mikkelcodes: > A couple of comments, but otherwise this looks good! Mainly I think we perhaps should refactor `tagbar` a bit as it's starting to get a bit big. At least an employee card should have its own component.

Thank you for the comments. I will have these things updated in a sec along with a few other changes.

packages/cms/schemas/tag.js
mikkelcodes: Makes sense

packages/cms/schemas/schema.js
```diff
@@ -21,7 +22,8 @@ export default createSchema({
     // in the studio.
     post,
     author,
-    category,
+    tag,
```
TheKnarf, line 25: Should this too be renamed to employee_tags?

packages/website/src/components/employeeCard/index.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { Shade } from '../icon';
+
+export const EmployeeCard = ({ data }) => {
+  const { image, id, experience, firstname, lastname, title } = data;
```
TheKnarf, line 5: It would probably be cleaner to have something like:

```
export const EmployeeCard = ({ image, id, experience, firstname, lastname, title }) => {
```

And then when using it one could use it as:

```
<EmployeeCard {...data} />
```

Or something similare.

packages/website/src/components/employeeCard/index.stories.js
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
+import { EmployeeCard } from './index';
+
+export default { title: 'components/employeeCard' };
+
+export const Default = () => <EmployeeCard />;
```
TheKnarf, line 6: Add some test data here?

packages/website/src/components/tagbar/index.js
```diff
@@ -1,84 +1,89 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { useEmployeeQuery } from '../../hooks/useEmployeeQuery';
 import { Tag } from '../icon';
```
TheKnarf, line 3: To make it slightly more clear that this is in-fact an icon we could import it as following:

```
```

and then use it as:

```
<Icon.Tag />
```

That makes it easier to understand when reading the code that it's an icon we're importing and using and not a tag in itself.

packages/website/src/components/tagbar/index.js
```diff
@@ -1,84 +1,89 @@
...
-  ];
-  const handleClick = (e) => {
-    const currentTag = e.target.firstElementChild.innerHTML;
-    if (tags.indexOf(currentTag) > -1) {
-      tags = tags.filter((tag) => {
-        return tag !== currentTag;
+export const Tags = () => {
+  const data = useEmployeeQuery();
+  const sanityTags = data.allSanityEmployeeTags.edges;
+  const sanityEmployees = data.allSanityEmployee.edges;
```
TheKnarf, line 9: Could the GraphQL stuff be moved out of the component itself and instead be ran as a part of the page? That way components inside `src/components` are somewhat `"pure"`, by which I mean that they don't do any data fetching. And only pages do fetch any data.

That also would mean that it would be easy to add anything from `src/components` to Storybook since we won't have to mock out the graphql / data fetching part.

packages/website/src/components/tagbar/index.js
```diff
@@ -1,84 +1,89 @@
...
-              onClick={handleClick}
-            >
-              <span className="transform -translate-y-1px pointer-events-none mx-auto">
-                {tag}
-              </span>
-            </button>
-          ))}
+          {sanityTags.map((sanityTag) => {
+            return (
+              <button
```
TheKnarf, line 66: Instead of using button here you could use the new `src/component/checkbox`-component.

TheKnarf:
I still have a few refactoring ideas that I commented on, but we're getting closer! :-)

packages/website/src/components/tagbar/index.js
```diff
@@ -1,84 +1,89 @@
...
-            <div>{employee.xp}</div>
-            <div>
-              <div>{employee.name}</div>
-              <div>{employee.title}</div>
-            </div>
-          </div>
-        ))}
+
+      {/* EMPLOYEE CARDS */}
+      <div className="grid gap-4 justify-center mx-auto sm:grid-cols-employees-sm nine:grid-cols-employees-nine 2xl:grid-cols-employees-lg">
```
TheKnarf, line 82: Maybe also take this bit out as a `Employees` component?

packages/website/src/components/tagbar/index.js
```diff
@@ -1,84 +1,89 @@
...
+
+  const filterEmployees = (tags) => {
+    const filteredEmployees = sanityEmployees.filter((employee) => {
+      let test;
+      employee.node.tags.forEach((node) => {
+        tags.indexOf(node.tag) > -1 ? (test = true) : (test = false);
+      });
+      return test;
+    });
+    return setEmployees(filteredEmployees);
```
TheKnarf, line 34: The `return` doesn't do anything here?

packages/website/src/components/tagbar/index.js
```diff
@@ -1,84 +1,89 @@
...
-      tags = tags.filter((tag) => {
-        return tag !== currentTag;
+export const Tags = () => {
+  const data = useEmployeeQuery();
+  const sanityTags = data.allSanityEmployeeTags.edges;
+  const sanityEmployees = data.allSanityEmployee.edges;
+
+  const [loaded, setLoaded] = useState(false);
+  const [tags, setTags] = useState([]);
+  const [employees, setEmployees] = useState(sanityEmployees);
```
TheKnarf, line 13: Maybe rename this from `employees` and to `filtered_employees`?

packages/website/src/components/tagbar/index.js
```diff
@@ -1,84 +1,89 @@
...
+
+  useEffect(() => {
+    if (!loaded) {
+      const arr = [];
+      sanityTags.forEach((sanityTag) => {
+        arr.push(sanityTag.node.tag);
```
TheKnarf, line 19: This bit can be rewritten as a map-function:

```
const arr = sanityTags.map(sanityTag => sanityTag.node.tag);
```

packages/website/src/components/tagbar/index.js
```diff
@@ -1,84 +1,89 @@
...
-    const currentTag = e.target.firstElementChild.innerHTML;
-    if (tags.indexOf(currentTag) > -1) {
-      tags = tags.filter((tag) => {
-        return tag !== currentTag;
+export const Tags = () => {
+  const data = useEmployeeQuery();
+  const sanityTags = data.allSanityEmployeeTags.edges;
+  const sanityEmployees = data.allSanityEmployee.edges;
+
+  const [loaded, setLoaded] = useState(false);
```
TheKnarf, line 11: This variable doesn't seem to be used anywhere except for in that one useEffect under, and therefore we should be able to actually just remove it. It doesn't seem to actually do anything. Instead under on line 16, we could just check for `if(sanityTags !== undefined && sanityTags !== null)` or something similar.

packages/website/src/components/tagbar/index.js
mikkelcodes: This is exactly what I meant to do. The query is intended for the whole employee page, hence the name useEmployeeQuery, and later, useBlogQuery, useArticleQuery, etc. Good catch..

packages/website/src/components/tagbar/index.js
mikkelcodes: Done

packages/website/src/components/tagbar/index.js
mikkelcodes: Return test? I just updated it btw. It returns the boolean of whether an employee's tag matches with the currently "activated" employee tags.

packages/website/src/components/tagbar/index.js
mikkelcodes: Totally.

packages/website/src/components/tagbar/index.js
mikkelcodes: Noted. Alright if I will update this with the next PR ?

TheKnarf (approved):

PR #66, opened 2020-11-15
Employee page

mikkelcodes:
I plan to refactor some components further, especially the tagbar component is a bit complex now and has room for improvement.


packages/website/src/hooks/useWindowDimensions.js
```diff
@@ -0,0 +1,27 @@
+import { useState, useEffect } from 'react';
+// import { window } from 'browser-monads';
```
mikkelcodes, line 2: If deployment to azure gives error due to window == undefined, try uncommenting this code.

TheKnarf: It does indeed look like you need to uncomment this to get the build working.

packages/website/src/layout.css
```diff
@@ -12,3 +12,11 @@
     }
   }
 }
+
+iframe .ytp-large-play-button,
+iframe .ytp-button {
+  display: none;
+}
+#movie_player > div.ytp-cued-thumbnail-overlay > button {
```
TheKnarf, line 20: What is this used for?

mikkelcodes: fixed

TheKnarf (approved):

PR #71, opened 2020-11-17
Switch fonts plugin

mikkelcodes:


mikkelcodes:
Redundant - this update is covered in a new PR

PR #77, opened 2020-11-22
Refactoring and nav component (desktop)

mikkelcodes:


packages/website/src/components/checkbox/index.js
```diff
@@ -0,0 +1,36 @@
+import React, { useState } from 'react';
+
+export const Checkbox = ({
+  children,
+  updateFilter,
+  defaultValue = undefined,
+  tag,
+}) => {
+  const [value, setValue] = useState(true);
```
TheKnarf, line 9: Did you have to refactor this to use internal state? It worked well as it was..

If you are going to change it to use internal state then you have to also update the unit tests and the storybook story.

packages/website/src/components/tagbar/index.js
```diff
@@ -1,9 +1,10 @@
 import React, { useEffect, useState } from 'react';
 import * as Icon from '../icon';
 import * as Button from '../button';
-import { EmployeeCard } from '../employeeCard';
-import { EmployeeBio } from '../employeeBio';
-import { Tag } from '../tag';
+// import { EmployeeCard } from '../employeeCard';
+// import { EmployeeBio } from '../employeeBio';
```
TheKnarf, line 5: You can remove the commented out imports if they're not in use.

packages/website/src/components/tagbar/index.js
```diff
@@ -125,45 +124,25 @@ export const Tags = ({ sanityTags, sanityEmployees }) => {
...
-              );
-            }
-            return null;
-          })
-        : null}
+      <EmployeeGroups
+        employeeGroups={employeeGroups}
+        activeBio={activeBio}
+        handleCardClick={handleCardClick}
+        handleCloseClick={handleCloseClick}
```
TheKnarf, line 144: Do we need `handleCardClick` and `handleCloseClick`? Isn't it just as well to leave those inside the component?

packages/website/src/pages/ansatte.js
```diff
@@ -0,0 +1,43 @@
...
+import { Description } from '../components/description';
+import { Tags } from '../components/tagbar';
+import { Cta } from '../components/cta';
+import { useEmployeeQuery } from '../hooks/useEmployeeQuery';
+
+const Employees = () => {
+  const data = useEmployeeQuery();
+  return (
+    <Layout>
+      <div className="bg-navy w-full pt-8 sm:pt-16 sm:pb-12 pb-4">
```
TheKnarf, line 13: Maybe this div could be made into a reusable component under `src/components`, something like `<Section.Navy>` or something?

mikkelcodes:
Just a side note, there is more refactoring to be done. I just thought to progress to the next component (navigation), and then I will come back and improve further

TheKnarf (commented):
You can either choose to update the unit tests and story for Checkbox, or you have to refactor it back to use an external state instead of an internal one as it originally was.

Otherwise, it looks mostly good!

TheKnarf (approved):

mikkelcodes:
Just did another round of refactoring. There is still another round left before "Tagbar" is in fact only a .. tagbar.

packages/website/src/components/tagbar/index.js
```diff
@@ -133,23 +62,8 @@ export const Tags = ({ sanityTags, sanityEmployees }) => {
...
-        <div />
-        <div
-          className="font-bold tracking-wider pr-2px"
-          onClick={handleViewMoreClick}
-        >
-          <Button.Line>Se Mer</Button.Line>
-        </div>
-      </div>
+      {/* CONTAINER WITH CARDS */}
+      <EmployeesContainer filteredEmployees={filteredEmployees} />
```
TheKnarf, line 66: The `EmployeesContainer`-component probably doesn't "care to know" that the employees are filtered, as such it makes more sense to call the property just `employees` so that it becomes `<EmployeeContainer employees={filteredEmployees} />`

packages/website/src/pages/ansatte.js
```diff
@@ -23,7 +23,7 @@ const Employees = () => {
           </span>
         </div>
         <div className="h-10 sm:h-16 md:h-24 mt-3" />
-        <Tags
+        <Tagbar
           sanityTags={data.allSanityEmployeeTag.edges}
```
TheKnarf, line 27: Similarly to how i commented over. The `tagbar`-component doesen't care that tags and employees comes from sanity, so it makes more sense with the following:

```
<Tagbar
  tags={data.allSanityEmployeeTag.edges}
  employees={data.allSanityEmployee.edges}
  />
```

TheKnarf (approved):

packages/website/src/pages/ansatte.js
mikkelcodes: This was already in the plans

packages/website/src/components/tagbar/index.js
mikkelcodes: I agree

PR #95, opened 2020-11-30
Blog overview page

mikkelcodes:


mikkelcodes:
Probably a few commits here and there that could be squashed

packages/website/gatsby-config.js
```diff
@@ -18,6 +18,13 @@ module.exports = {
         overlayDrafts: true,
       },
     },
+    {
+      resolve: `gatsby-source-filesystem`,
```
TheKnarf, line 22: What is this needed for?

TheKnarf:
Build error:

```
Can't resolve '../blogTagbar' in '/home/runner/work/alv-website/alv-website/packages/website/src/components/blogSection'
```

The import is importing `blogTagbar` but the folder is called `blogTagBar`, I recommend renaming it to `blogTagbar`.

packages/website/src/components/tagbar/index.js
```diff
@@ -1,12 +1,13 @@
 import React, { useState } from 'react';
 import * as Icon from '../icon';
-import { EmployeesContainer } from '../employeesContainer';
+import { GridContainer } from '../gridContainer';
 import { Checkbox } from '../checkbox';
 
 export const Tagbar = ({ sanityTags, sanityEmployees }) => {
-  const [filteredEmployees, setFilteredEmployees] = useState(
+  const [filteredContent, setFilteredContent] = useState(
```
TheKnarf, line 7: Not sure why this was renamed? I think `filteredEmployees` made more sense here than just content?

packages/website/src/components/tagbar/index.js
```diff
@@ -63,7 +64,7 @@ export const Tagbar = ({ sanityTags, sanityEmployees }) => {
         </div>
       </div>
       {/* CONTAINER WITH CARDS */}
-      <EmployeesContainer filteredEmployees={filteredEmployees} />
+      <GridContainer filteredContent={filteredContent} />
```
TheKnarf, line 67: Are we reusing the GridContainer other places?

packages/website/gatsby-config.js
mikkelcodes: This is used to source a fluid fallback image for whenever we need fallback images

TheKnarf: I'm confused, didn't that already work without that plugin?

packages/website/src/components/tagbar/index.js
mikkelcodes: The idea was that the same component will be used for filtering blog posts as well, so I changed to a more generic name

packages/website/src/components/tagbar/index.js
mikkelcodes: Yes, the idea behind this change is the same as the above. Reusing the container

packages/website/gatsby-config.js
mikkelcodes: Yes exactly, this is for use with gatsby-image

mikkelcodes:
> Build error:
> 
> ```
> Can't resolve '../blogTagbar' in '/home/runner/work/alv-website/alv-website/packages/website/src/components/blogSection'
> ```
> 
> The import is importing `blogTagbar` but the folder is called `blogTagBar`, I recommend renaming it to `blogTagbar`.

Good recommendation

TheKnarf (approved):

PR #204, opened 2021-02-03
Bux

mikkelcodes:


packages/website/src/components/employeeBio/index.js
```diff
@@ -9,22 +9,22 @@ export const EmployeeBio = ({
...
-    .concat(lastname.split(' '))
-    .map((name) => name.toLowerCase());
-  employeeSlug = employeeSlug.join('-');
-  const videoUrlWebm = video ? video.asset.url : false;
-  const videoUrlYoutube = videoLink || false;
+  const [webM, setWebM] = useState(null);
+  const [youtube, setYoutube] = useState(videoLink || false);
+  useEffect(() => {
+    setWebM(video ? video.asset.url : false);
+    setYoutube(videoLink || false);
```
TheKnarf, line 24: Replace useState and useEffect with a useMemo

packages/website/src/components/employeeBio/index.js
```diff
@@ -56,26 +56,27 @@ export const EmployeeBio = ({
               </div>
               {/* ---- Above: Visible on Mobile only ---- */}
               <div className="sm:h-80 sm:w-140 mb-8 sm:mb-5">
-                {videoUrlWebm && !isIOS && !isSafari && (
+                {webM && !isIOS && !isSafari && (
```
TheKnarf, line 59: Move this logic up into a single variable, then replace this with a switch statement.
Something like:

```
const supportWebM = !isOS && !isSafari;
const type = (supportWebM && webM) ? 'webm' :
                      videoUrlYoutube ? 'youtube' :
                    'fallback';
```

packages/website/src/pages/ansatte.js
```diff
@@ -8,7 +8,27 @@ import { useEmployeeQuery } from '../hooks/useEmployeeQuery';
...
-  const activeCard = location.state ? location.state.activeCard : null;
+  const allEmployees = data.allSanityEmployee.edges.map((el) => {
+    const employee = el.node;
+    let slug =
+      employee.firstname &&
+      employee.firstname.split(' ').concat(employee.lastname.split(' '));
+    slug = '#' + slug.join('-').toLowerCase();
+    employee.slug = slug;
+    return employee;
+  });
```
TheKnarf, line 19: Perhaps move this logic into the hook

packages/website/src/pages/ansatte.js
```diff
@@ -8,7 +8,27 @@ import { useEmployeeQuery } from '../hooks/useEmployeeQuery';
...
+  if (location.state) {
+    activeCard = location.state.activeCard;
+  } else {
+    if (location.hash) {
+      const card = allEmployees.find((el) => {
+        return el.slug === location.hash;
+      });
+      activeCard = card.id;
+    }
+  }
```
TheKnarf, line 31: Move all of this logic into a useMemo

TheKnarf (changes requested):

packages/website/src/components/employeeBio/index.js
mikkelcodes: Please suggest a syntax here

packages/website/src/pages/ansatte.js
mikkelcodes: Planned for the broader hook update (object destructuring etc)

packages/website/src/components/employeeBio/index.js
mikkelcodes: I guess type will have to live inside useMemo/useState as it must be updated with every employee card click from the same group/row

packages/website/src/pages/ansatte.js
mikkelcodes: Sure. Could you just enlighten me abt the purpose of this? Maybe there is something I have overlooked

packages/website/src/pages/ansatte.js
mikkelcodes: nvm, updated this

packages/website/src/pages/ansatte.js
TheKnarf: Use memo will cache that logic, and then you can simply add location as a dependency to make it clear for people who read the code when it would update. It's not strictly necessary here but I think it would help a bit. It also means that `activeCard` won't need to use `let`.

packages/website/src/components/employeeBio/index.js
TheKnarf: When you have a useState thats only updated from a useEffect you can always combine them into a useMemo.

packages/website/gatsby-node.js
```diff
@@ -49,13 +49,32 @@ if (process.env.YT_API) {
...
           return item.snippet;
         });
+        if (list.pageInfo.totalResults > 50) {
+          let nextPageToken = list.nextPageToken;
+          while (nextPageToken) {
+            const apiCall = await fetch(
+              `https://youtube.googleapis.com/youtube/v3/playlistItems?part=snippet&maxResults=50&playlistId=${el.id}&pageToken=${nextPageToken}&key=${process.env.YT_API}`
+            );
+            const nextPageList = await apiCall.json();
+            const nextPageItems = await nextPageList.items.map((item) => {
```
TheKnarf, line 66: The inside of the map is now duplicated between this line and line 52 and could probably be extracted into a function.

packages/website/gatsby-browser.js
```diff
@@ -16,4 +16,7 @@ export const onRouteUpdate = ({ location }) => {
   if (location.state) {
     window.setTimeout(scrollTo(`#${location.state.employee}`), 100);
   }
+  if (location.hash) {
+    window.setTimeout(scrollTo(location.hash), 500);
```
TheKnarf, line 20: Is this and line 17 both needed?

packages/website/src/components/employeeSection/index.js
```diff
@@ -8,22 +8,20 @@ export const EmployeeSection = ({
...
     const filteredEmployeeArr = allEmployees.filter((employee) => {
       let test;
-      employee.node.tags.forEach((node) =>
+      employee.tags.forEach((node) =>
         !test ? (test = tags.indexOf(node.tag) > -1) : null
       );
       return test;
     });
-    return setFilteredContent(filteredEmployeeArr.map((el) => el.node));
+    return setFilteredContent(filteredEmployeeArr);
```
TheKnarf, line 24: Perhaps remove `return` here as it doesn't seem to do anything?

packages/website/src/components/employeeSection/index.js
```diff
@@ -8,22 +8,20 @@ export const EmployeeSection = ({
...
 
   const [tags, setTags] = useState(allTags.map((tag) => tag.node.tag));
 
   // Filter employees based on active tags
   const filterEmployees = (tags) => {
     const filteredEmployeeArr = allEmployees.filter((employee) => {
       let test;
-      employee.node.tags.forEach((node) =>
+      employee.tags.forEach((node) =>
         !test ? (test = tags.indexOf(node.tag) > -1) : null
```
TheKnarf, line 20: Could this be refactored into `employee.tags.filter` instead of `forEach`?

packages/website/src/components/employeeBio/index.js
```diff
@@ -9,22 +9,35 @@ export const EmployeeBio = ({
...
-    .map((name) => name.toLowerCase());
-  employeeSlug = employeeSlug.join('-');
-  const videoUrlWebm = video ? video.asset.url : false;
-  const videoUrlYoutube = videoLink || false;
+  const webM = useMemo(() => {
+    return video ? video.asset.url : false;
+  }, [video]);
+  const youtube = useMemo(() => {
+    return videoLink || false;
+  }, [videoLink]);
```
TheKnarf, line 25: Since `const webM` and `const youtube` is only used inside the last useMemo it's better to combine all of this into a single useMemo (that way React doesn't have to calculate and keep track of 3 useMemo's when it's not needed).

TheKnarf (approved):

PR #213, opened 2021-02-11
Bux

mikkelcodes:


TheKnarf (approved):

packages/website/gatsby-node.js
```diff
@@ -34,7 +53,29 @@ exports.createPages = async ({ graphql, actions }) => {
...
+      path: `/vi-tilbyr/${edge.node.slug.current}`,
+      context: {
+        slug: edge.node.slug.current,
+      },
+    });
+  });
+
+  // Create redirects
+  res.data.allSanityRedirect.edges.forEach((redirect) => {
+    createRedirect({
```
TheKnarf, line 70: How does redirecting work in Gatsby? Do you have a link to some documentation?

packages/website/src/templates/service.js
```diff
@@ -0,0 +1,109 @@
+import React from 'react';
+import Layout from '../layout';
+import { graphql } from 'gatsby';
+import { ServiceNavList } from '../components/serviceNavList';
+import { RolesList } from '../components/rolesList';
+import { NavyIntroImage } from '../components/navyIntroImage';
+import { CtaSection } from '../components/ctaSection';
+
+const Systemutvikling = ({ data }) => {
```
TheKnarf, line 9: Is probably supposed to be named `const Service` or something similare?

packages/website/gatsby-node.js
TheKnarf: This may be used in the future. https://www.gatsbyjs.com/plugins/gatsby-plugin-nginx-redirect/?=redirect

But for now, let's just remove until we can set it all up.

TheKnarf (dismissed):

TheKnarf (approved):

PR #250, opened 2021-03-01
Seo

mikkelcodes:


packages/cms/deskStructure.js
```diff
@@ -11,13 +11,79 @@ export default () =>
         ),
       S.divider(),
       S.listItem()
-        .title('Landing Page')
+        .title('Pages (static gen)')
```
TheKnarf, line 14: Will this show as `Pages (static gen)` in Sanity? If so it might just as well be just `Pages`?

packages/cms/schemas/categoryPage.js
```diff
@@ -1,8 +1,25 @@
 export default {
   name: 'categoryPage',
-  title: 'Category Pages',
+  title: 'Category Pages (dynamic gen)',
```
TheKnarf, line 3: I'm unusre why we would write `(dynamic gen)` here. How does that help a non-developer that uses Sanity to change something?

packages/cms/deskStructure.js
mikkelcodes: Yes, I wrote it to differentiate between statically generated pages and dynamically generated (dynamic gen) pages. I should remove the descriptions inside () ?

packages/cms/schemas/categoryPage.js
mikkelcodes: Yes, depends on the tech savviness I guess. I would have wanted it there so that's why I set it like that. But I will remove it

packages/cms/schemas/servicePage.js
```diff
@@ -1,8 +1,25 @@
 export default {
   name: 'services',
-  title: 'Service Pages',
+  title: 'Service Pages (dynamic gen)',
```
TheKnarf, line 3: I still don't think `dynamic gen` means anything for an enduser.

TheKnarf (approved):

TheKnarf (approved):

PR #265, opened 2021-03-15
Update gatsby

mikkelcodes:


TheKnarf (changes requested):
If this is supposed to update gatsby then you're missing changes in `package.json` and `yarn.lock`

mikkelcodes:
> If this is supposed to update gatsby then you're missing changes in `package.json` and `yarn.lock`

Looks like I am still getting the fsevents error. I thought the upgrade went through but it failed even after doing a yarn fsevents rebuild (successful) followed by yarn upgrade-interactive (fails). 

For the sake of progressing, I propose we put a pin in this upgrade and see if other patches/minor updates come along and makes the transition to 3.0 smooth for all systems (including m1 macs..). What do you say?

TheKnarf:
You should be able to just ignore errors from fsevent, if it can't install it it should automatically fallback to another solution. So if everything except fsevent works then we should be able to go forward with the upgrade.

mikkelcodes:
> You should be able to just ignore errors from fsevent, if it can't install it it should automatically fallback to another solution. So if everything except fsevent works then we should be able to go forward with the upgrade.
> 
> What kind of problems do you get from `yarn upgrade-interactive`?

Screenshot 2021-03-15 at 15 51 03

mikkelcodes:
I was able to make some progress and install gatsby 3.0 

Screenshot 2021-03-16 at 08 25 42

but compilation now fails with this error: 

Screenshot 2021-03-16 at 08 25 13

TheKnarf:
Hm weird, feel free to push up the changes to this branch. Then either I can look over it if I got some time, or we can let it lie her for a few weeks and try again later.

PR #266, opened 2021-03-17
Video episode pages with url

boye-s:
Each video now has its own url. Urls are based on running the video title through [slugify](https://yarnpkg.com/package/slugify).

Known issues:
* The YouTube API is called twice. Once for create page and once for creating the YouTube video graphql query. It needs to be hung up to DRY.
* Breadcrumbs look kinda weird being based on the slugified video titles.
* State variables included in `<Link />` are probably unnecessary now.


package.json
```diff
@@ -28,5 +28,8 @@
       "pre-commit": "run-s -s lint"
     }
   },
-  "prettier": "@blueberrypie/prettier-config"
+  "prettier": "@blueberrypie/prettier-config",
+  "dependencies": {
+    "slugify": "^1.4.7"
```
TheKnarf, line 33: Can you move this into the `package.json` thats inside `packages/website`? We don't need this as a global dependency, it should be enough for it to be a dependency for the website.

packages/website/gatsby-node.js
```diff
@@ -47,6 +50,72 @@ exports.createPages = async ({ graphql, actions }) => {
       }
     `
   );
+  console.log(process.env);
```
TheKnarf, line 53: This can be removed

TheKnarf:
I haven't cloned it locally and tested it yet, but just looking over the code it looks really good. See if you can find time to fix the two comments over.

packages/website/gatsby-node.js
```diff
@@ -59,6 +62,72 @@ exports.createPages = async ({ graphql, actions }) => {
...
+      playlists.forEach((playlist) => {
+        playlist.forEach((video) => {
+          const slug = slugify(video.title.replace(' |', ''), {
+            remove: /[*+~.()|#'"!:@]/,
+            lower: true,
+          });
+
+          createPage({
+            component: videoTemplate,
+            path: `/videoserie/${slug}`,
```
TheKnarf, line 121: Is there anyway we could also get a slug for the playlist? I would like the url to be something like `/videoserie/${playlistSlug}/${slug}`.

packages/website/src/templates/video.js
```diff
@@ -0,0 +1,37 @@
...
+const VideoTemplate = ({ location, pageContext }) => {
+  const video = pageContext.video;
+  const data = useVideoseriesQuery();
+  const [playlist, setPlaylist] = useState(data.playlists[0]);
+
+  useState(() => {
+    location.state && setPlaylist(location.state.playlist);
+  }, []);
+
+  //
```
TheKnarf, line 17: Comment can be removed

packages/website/src/templates/video.js
```diff
@@ -0,0 +1,37 @@
...
+import { VideoEpisode } from '../components/videoEpisode';
+import { useVideoseriesQuery } from '../hooks/useVideoseriesQuery';
+
+const VideoTemplate = ({ location, pageContext }) => {
+  const video = pageContext.video;
+  const data = useVideoseriesQuery();
+  const [playlist, setPlaylist] = useState(data.playlists[0]);
+
+  useState(() => {
+    location.state && setPlaylist(location.state.playlist);
```
TheKnarf, line 14: This whole `useState` now seem unnessesary since we get one html-page per video..

packages/website/src/templates/video.js
```diff
@@ -0,0 +1,37 @@
...
+import Layout from '../layout';
+import { EpisodeCards } from '../components/episodeCards';
+import { SocialShare } from '../components/socialShare';
+import { VideoEpisode } from '../components/videoEpisode';
+import { useVideoseriesQuery } from '../hooks/useVideoseriesQuery';
+
+const VideoTemplate = ({ location, pageContext }) => {
+  const video = pageContext.video;
+  const data = useVideoseriesQuery();
+  const [playlist, setPlaylist] = useState(data.playlists[0]);
```
TheKnarf, line 11: You can probably just use `data.playlists[0]` directly instead of saving it in a `useState`now that we have a templated page.

TheKnarf (approved):

PR #268, opened 2021-03-18
add job opening pages backend and frontend

mikkelcodes:


packages/website/src/templates/career.js
```diff
@@ -0,0 +1,74 @@
...
+import PortableText from '@sanity/block-content-to-react';
+import Layout from '../layout';
+import styles from './Article.module.css';
+import { graphql } from 'gatsby';
+import { NavyIntro } from '../components/navyIntro';
+
+const Career = ({ data }) => {
+  const { sanityOpenPostionPage } = data;
+
+  console.log(sanityOpenPostionPage);
```
TheKnarf, line 11: console.log can be removed

packages/website/src/templates/career.js
```diff
@@ -0,0 +1,74 @@
...
+          <PortableText
+            blocks={sanityOpenPostionPage._rawJobDescription}
+            projectId="mnr37rl0"
+            dataset="production"
+          />
+        </span>
+        <div>
+          <div
+            className="lg:mt-10 mt-20"
+            dangerouslySetInnerHTML={{ __html: sanityOpenPostionPage.embed }}
```
TheKnarf, line 42: Is there no better way to do this that doesn't include a dangerouslySetInnerHTML?

mikkelcodes: I'm looking into a substitute for this

mikkelcodes: If we can only source a link, that would make it easier. However, the current embed includes both a script tag and an iframe. But I'll try to come up with something

TheKnarf: Allright, perhaps we just merg it then and look for a solution in the future. I would have guessed that perhaps Sanity had some official React package with components for embeding stuff from them, but if not then maybe this is the best solution.

TheKnarf (approved):

PR #300, opened 2021-04-07
fix bux

mikkelcodes:


packages/website/gatsby-config.js
```diff
@@ -73,17 +73,6 @@ module.exports = {
         },
       },
     },
-    {
-      resolve: 'gatsby-plugin-google-tagmanager',
```
TheKnarf, line 77: Why is this removed?

packages/website/src/components/checkbox/index.js
```diff
@@ -1,7 +1,7 @@
 import React, { useState } from 'react';
 
 export const Checkbox = ({ children, onChange, tag, defaultChecked }) => {
-  if (defaultChecked === undefined) defaultChecked = true;
+  if (defaultChecked === undefined) defaultChecked = false;
```
TheKnarf, line 4: I don't think that we should keep changing this file back and forth. We have changed the defaultChecked value several times now.... Maybe just change how you use the component instead of changing this file? Every time you change it you also change the test / have to change the test; and the point of unit tests are to test that the code work as intended, not to keep changing it. If we have a unit test for something that is intended to mean that this component work as intended.

So change the use of the component not the component itself.

packages/website/src/components/employeeSection/index.js
```diff
@@ -24,11 +24,17 @@ export const EmployeeSection = ({
     return setFilteredContent(filteredEmployeeArr);
   };
 
+  useEffect(() => {
```
TheKnarf, line 27: This can probably be refactored into using an `useMemo`?

packages/website/src/components/imageTextFull/index.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import styles from '../../templates/Article.module.css';
+import styles from '../../templates/Service.module.css';
```
TheKnarf, line 2: This component should probably have it's own css-module instead of importing a css module from somewhere else? If you need some shared style then refactor that into it's own component or use css variables.

packages/website/src/components/navigation/index.js
```diff
@@ -63,7 +63,7 @@ export const Navigation = ({
             Vi Tilbyr
           </LargeLink>
           {categoryPages &&
-            categoryPages.slice(0, 3).map((categoryPage) => {
+            categoryPages.slice(0, 2).map((categoryPage) => {
```
TheKnarf, line 66: Why are we slicing to `2` here? Maybe take that out into a variable so that it's easier to read the code instead of using a "magic number".

packages/website/src/components/navigation/index.js
```diff
@@ -113,7 +113,7 @@ export const Navigation = ({
         >
           <div className="eight:mt-20 eight:pt-7 xl:pt-10" />
           {categoryPages &&
-            categoryPages.slice(3, categoryPages.length).map((categoryPage) => {
+            categoryPages.slice(2, categoryPages.length).map((categoryPage) => {
```
TheKnarf, line 116: Why are we slicing to `3` here? Maybe take that out into a variable so that it's easier to read the code instead of using a "magic number".

packages/website/src/components/navyIntroImage/index.js
```diff
@@ -11,8 +11,16 @@ export const NavyIntroImage = ({ title, description, image, internalLink }) => (
       style={{ gridTemplateColumns: '60% auto' }}
     >
       <div className="transform sm:-translate-x-5 sm:px-0 px-12 flex sm:block justify-center items-center flex-col">
-        <Title align="center twelve:text-left text-blog sm:text-about-xl">
-          {title}
+        <Title align={`center twelve:text-left sm:text-about-xl`}>
+          <span
+            className={`${
+              title.length > 15
```
TheKnarf, line 17: Why are we switching on `title.length` here? That seems weird. At least take the `title.length > 15` portion out into its own variable so that someone who reads the code knows what it is.

packages/website/src/pages/vi-tilbyr.js
```diff
@@ -41,33 +41,70 @@ const Services = () => {
...
+          </div>
+          <div className="bg-servicesgray text-navy px-5 sm:px-12 overflow-hidden">
+            <ServicesSection
+              description={section3description}
+              title={data.sanityOurServicesPage.section3link.heroHeading}
+              link={data.sanityOurServicesPage.section3link.slug.current}
+            >
+              {cards
+                .filter(
+                  (card) => card.parentPage.slug.current === 'systemutvikling'
```
TheKnarf, line 69: Why are we filtering on `systemutvikling` here? We probably shouldn't have that hardcoded. This feels like something that should live in Sanity.

packages/website/src/templates/Service.module.css
```diff
@@ -0,0 +1,115 @@
+.mainImage {
```
TheKnarf, line 1: Why is all of this a css-module and not just tailwindcss?

TheKnarf (changes requested):
I'd like to either get clearance in the issues over, or have them fixed.

packages/website/gatsby-config.js
mikkelcodes: To my understanding, we're only using the Google Analytics part and a separate google analytics plugin already takes care of that embed. Furthermore, it started throwing strange errors here and there although it wasn't really doing anything. Removed it as a quick fix and we can look to reintroduce it later if necessary

packages/website/src/components/checkbox/index.js
mikkelcodes: Yep, that makes sense

packages/website/src/components/imageTextFull/index.js
mikkelcodes: I'll create an issue to define a global blockcontent (from sanity) css stylesheet.

packages/website/src/components/navigation/index.js
mikkelcodes: Yes - good one. When I think about it, this should be improved programmatically - the number should be determined based on how many child pages each category have.

packages/website/src/components/navigation/index.js
mikkelcodes: Same as above

packages/website/src/components/navyIntroImage/index.js
mikkelcodes: I discussed this issue yesterday with Jakob and we decided we'll do a separate field in Sanity to replace this functionality with something more flexible. It's about long words that need to scale down to fit in the view. As per mobile design spec, the font size of category/service page titles are way too large for several titles containing 20+ characters

TheKnarf (dismissed):
Approving the PR with the assumption that you have made issues out of the things we have discussed.

packages/website/src/templates/Service.module.css
mikkelcodes: Because we can't reach jsx-elements generated by the PortableText component directly with tailwind. PortableText is a package from sanity that converts Rich Text into jsx-elements.

mikkelcodes: I will consolidate a few of the template page CSS stylesheets. Made an issue on it

TheKnarf (approved):

PR #305, opened 2021-04-10
nav update

mikkelcodes:


packages/website/src/components/navItems/index copy.js
```diff
@@ -0,0 +1,94 @@
+import React, { useState } from 'react';
```
TheKnarf, line 1: Why do you have a file called `index copy.js`?  Either delete it or rename it.

packages/website/src/components/navItems/index.js
```diff
@@ -4,7 +4,7 @@ import Link from 'gatsby-link';
 import styles from './navItems.module.css';
 
 const ConditionalLinkWrapper = ({ children, link }) =>
-  link !== undefined ? <Link to={link}>{children}</Link> : children;
+  link !== false ? <Link to={link}>{children}</Link> : children;
```
TheKnarf, line 7: Maybe its better just to test it like:

```
link ? <Linke to={link}>{children}<Link> : children
```

packages/website/src/components/navItems/index.js
```diff
@@ -13,35 +13,35 @@ export const LargeLink = ({ children, margin, link }) => {
   };
   return (
     <>
-      <ConditionalLinkWrapper condition={link !== undefined} link={link}>
-        {link === undefined ? (
+      <ConditionalLinkWrapper condition={link !== false} link={link}>
```
TheKnarf, line 16: Why do we have a `ConditionalLinkWrapper` and then also a bunch of if-statements inside it? Maybe this code would be easier to read if the component simply had one big if-statement and two return statements?

packages/website/src/components/navigation/index old.js
```diff
@@ -0,0 +1,180 @@
+import React from 'react';
```
TheKnarf, line 1: Why do we have a file called `index old.js`? Either rename it or delete it.

packages/website/src/components/navigation/index.js
```diff
@@ -13,175 +13,186 @@ export const Navigation = ({
...
+        <div className="eight:flex justify-center relative z-40 mx-auto auto eight:mb-0 eight:mb-25 eight:mx-auto max-w-1200 items-center mb-12 eight:mb-0">
+          <div className="">
+            <Link to="/">
+              <Logo.White />
+            </Link>
+          </div>
+          <div className="absolute right-0 transform -translate-x-20 eight:flex hidden">
             <a
-              href="https://twitter.com/alvnoas/"
+              href="https://www.instagram.com/alvnoas/"
```
TheKnarf, line 30: Move the social links into it's own component. I think you even already have a shared component for this?

packages/website/src/pages/index.js
```diff
@@ -15,6 +16,7 @@ const Index = ({ data, location }) => {
     { line: 'konsulentselskap', classes: '', dot: true },
   ];
 
+  const { width } = useWindowDimensions();
```
TheKnarf, line 19: Is there any reason to call this here and send it down instead of just using it in the component where it's needed?

TheKnarf (changes requested):
Fix the linting that fails, and see comments above.

packages/website/src/components/ourServices/index.js
```diff
@@ -54,7 +50,7 @@ export const OurServices = ({ image }) => {
             >
               <p className="uppercase pointer-events-none">Prosjektleder</p>
               {activeItem === '1' && (
-                <ListContent link={dummyLink}>
+                <ListContent link="/vi-tilbyr/prosjektledelse">
```
TheKnarf, line 53: This needs to be dynamic based on what's in Sanity.

TheKnarf (approved):

Issue #301, opened 2021-04-08
Responsiv bug


jakobhg:
Bug on responsiveness on "Jobbe i Alv" -> job posts

![Screenshot 2021-04-14 at 16 21 30](https://user-images.githubusercontent.com/213777/114726245-7ff10580-9d3d-11eb-82d7-a2ca488abeec.png)

TheKnarf:
This happens if the height of the browser window is to small.

mikkelcodes:
fixed in #317

Issue #307, opened 2021-04-12
vi tilbyr, graphical error


professoralv:
split the text to resolve issue
![image](https://user-images.githubusercontent.com/73942505/114421760-ef3bed80-9bb5-11eb-83a5-0351ab8605c9.png)

mikkelcodes:
fixed in pr #305

PR #317, opened 2021-04-14
update long word formatting - service card

mikkelcodes:


packages/website/src/components/servicesCard/index.js
```diff
@@ -40,3 +22,24 @@ export const ServicesCard = ({
 };
 
 export default ServicesCard;
+
+const SplitWord = ({ word }) => {
```
TheKnarf, line 26: Is this really needed? Isn't it enough to do `text-overflow: ellipsis` or something similare?

packages/website/src/components/servicesCard/index.js
```diff
@@ -40,3 +22,24 @@ export const ServicesCard = ({
...
+  const longWord = 15;
+  let formattedWord;
+
+  if (word.includes('-') && word.length > longWord)
+    formattedWord = word.split('-');
+
+  return (
+    <>
+      {formattedWord &&
```
TheKnarf, line 35: Just use `if` and have two return statements...

TheKnarf (changes requested):

packages/website/src/components/servicesCard/index.js
mikkelcodes: I think it is :/ These are titles like "systemutviklingsledelse" that are split into:
"systemutviklings-
ledelse"
and ending it with "systemutviklings..." might not be practical. But the method you proposed could be applied to blog post card descriptions to simplify things.

TheKnarf: Css hypens then? https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens

mikkelcodes: Screenshot 2021-04-14 at 19 40 17

Is this an acceptable look? (css hyphens)

TheKnarf: Seems fine for me.

packages/website/src/components/servicesCard/index.js
```diff
@@ -40,3 +22,20 @@ export const ServicesCard = ({
 };
 
 export default ServicesCard;
+
+const SplitWord = ({ titleSplit, title }) => (
```
TheKnarf, line 26: Remove unused component.

TheKnarf (changes requested):

packages/website/src/components/navyIntro/index.js
```diff
@@ -3,23 +3,16 @@ import BackgroundImage from 'gatsby-background-image-es5';
 import { Title } from '../title';
 import { Description } from '../description';
 
-export const NavyIntro = ({ title, description, image, headingSplit }) => (
+export const NavyIntro = ({ title, description, image }) => (
   <>
     {image ? (
       <div className="sm:z-40 -mt-35 -mb-2">
         <BackgroundImage fluid={image}>
-          <div className="bg-navy bg-opacity-75 w-full pt-10 sm:h-50vh h-screen flex flex-col justify-center items-center sm:pt-16 sm:pb-16 pb-4">
```
mikkelcodes, line 11: This line is from and resolves issue #301 but the same file also contains fixes for #307 (wordwrap)

TheKnarf (approved):

PR #327, opened 2021-04-16
Upgrade yarn

mikkelcodes:


packages/website/package.json
```diff
@@ -28,7 +30,7 @@
     "react-slick": "^0.28.0",
     "react-three-fiber": "^5.3.17",
     "slugify": "^1.4.7",
-    "tailwindcss": "^1.9.6",
+    "tailwindcss": "latest",
```
TheKnarf, line 33: Weird, it says latest instead of the actual version. That seems like a yarn bug.

TheKnarf (approved):

PR #342, opened 2021-04-22
add also read to articles

mikkelcodes:


packages/website/src/components/alsoRead/index.js
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import Link from 'gatsby-link';
+import { PostCard } from '../postCard';
+import * as Icon from '../icon';
+
+export const AlsoRead = ({ articles, currentTags, currentAuthor }) => {
+  const sortedArticles = articles.edges
```
TheKnarf, line 7: Could the sorting be refactored into a function? That way we could write some Unit tests for it to ensure that it works like we want it to.

PR #371, opened 2021-04-29
Adding scheduling for prod deploys.

TheKnarf:


.AzureDevOps/pipelines/azure-pipelines.yml
```diff
@@ -4,17 +4,19 @@ trigger:
...
-#  branches:
-#    include:
-#    - prod
+schedules:
+- cron: "30 10 * * 2"
+  displayName: Thusdays at 1030
+  branches:
+    include:
+    - prod
+  always: true
```
TheKnarf, line 13: Tested it in an isolated pipeline and it seemd that if I add `always: true` that scheduling now works.

Trulsaa: To force a pipeline to run even when there are no code changes, you can use the always keyword. 👍

Trulsaa (approved):

PR #375, opened 2021-05-05
Issue 205

mikkelcodes:


packages/website/src/components/employeeGroup/index.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import { EmployeeCard } from '../employeeCard';
 import { EmployeeBio } from '../employeeBio';
+// import window from 'browser-monads';
```
TheKnarf, line 4: Maybe remove outcommented code?

packages/website/src/components/featuredTeam/index.js
```diff
@@ -3,11 +3,38 @@ import { EmployeeCard } from '../employeeCard';
...
 import { Title } from '../title';
 import * as Button from '../button';
+import slugify from 'slugify';
 
 export const FeaturedTeam = ({ team, notitle, fallbackImg, notransparent }) => {
   const handleClick = (e) => {
     return e;
   };
+
+  // let authorSlug;
```
TheKnarf, line 13: Maybe remove outcommented code?

packages/website/src/components/featuredTeam/index.js
```diff
@@ -3,11 +3,38 @@ import { EmployeeCard } from '../employeeCard';
...
+  // }
+
+  const linkableTeam = team.map((el) => {
+    let slug;
+    if (el.firstname && el.lastname) {
+      slug = el.firstname
+        .split(' ')
+        .concat(el.lastname.split(' '))
+        .map((name) => name.toLowerCase());
+      slug = slugify(slug.join('-'));
```
TheKnarf, line 33: Wouldn't the following be enough?

```
slug = slugify(`${el.firstname} ${el.lastname}`.toLowerCase());
```

packages/website/src/hooks/useEmployeeQuery.js
```diff
@@ -79,7 +80,7 @@ export const useEmployeeQuery = () => {
       employee.firstname &&
       employee.firstname.split(' ').concat(employee.lastname.split(' '));
     slug = '#' + slug.join('-').toLowerCase();
-    employee.slug = slug;
+    employee.slug = slugify(slug);
```
TheKnarf, line 83: If your copy pasting the logic for how to make employee slugs in two places, maybe it could be refactored into a common function? No need to redo this logic both here and in the `featuredTeam` component. Better if both include a file with a function that does it the same way for both. That way if we have to change it in the future we only have to update one file. It would also make it easier to write a Jest unit test on it. I would love it if added a test for this.

packages/website/src/pages/ansatte.js
```diff
@@ -17,13 +18,11 @@ const Employees = ({ location }) => {
   let activeCard;
   if (location.state) {
```
TheKnarf, line 19: Do we need to check `location.state` or can we always use `location.hash`?

packages/website/src/templates/article.js
```diff
@@ -46,7 +47,7 @@ const ArticleTemplate = (props) => {
       .split(' ')
       .concat(postAuthor.lastname.split(' '))
       .map((name) => name.toLowerCase());
-    authorSlug = authorSlug.join('-');
+    authorSlug = slugify(authorSlug.join('-'));
```
TheKnarf, line 50: Again it seems that we keep copy-pasting the logic for employee slugs. Better to have a function in a common file that makes employee slugs in the same way.

packages/website/src/components/featuredTeam/index.js
mikkelcodes: absolutely

TheKnarf (changes requested):
TODO before approving:

- Merge inn the other PR's and then update and rebase this branch (`git fetch origin && git rebase origin/master`)

- Fix the comments above, the most important one is a common function for creating slugs from an employee

- Write a Jest unit test to test the new `createSlugForEmployee` function.

TheKnarf (approved):

PR #396, opened 2021-05-11
Issue 333

mikkelcodes:


packages/website/src/components/serializers/Serializers.module.css
```diff
@@ -0,0 +1,68 @@
+.wrapper {
```
mikkelcodes, line 1: This is styled in CSS modules because Tailwind selectors use lower specificity CSS modules, which already provide styles for rich text from Sanity.

packages/website/src/components/serializers/index.js
```diff
@@ -0,0 +1,96 @@
+import React from 'react';
```
TheKnarf, line 1: The name `serializers` is not obvius for me what it's supposed to be?

packages/cms/schemas/blockContent.js
```diff
@@ -64,5 +64,9 @@ export default {
       type: 'image',
       options: { hotspot: true },
     },
+    {
+      title: 'CTA',
```
TheKnarf, line 68: Should this be something like blogpostCTA or something?

mikkelcodes: I'll make this work universally for all sanity rich text fields. I think that can be valuable and it's a quick update. With that update, this name makes more sense also

TheKnarf (approved):

PR #401, opened 2021-05-21
Add script files that downloads a backup of the sanity production data

Trulsaa:
Should be further developede to add the backup to some storage

### DIFF
--- a/packages/scripts/backup.sh
+++ b/packages/scripts/backup.sh
@@ -1,6 +1,4 @@
-#!/usr/bin/env bash
-
-set -e
+#!/usr/bin/env -S bash -e
 
 KEY_VAULT="alvnoprodkv"
 SUBSCRIPTION="AlvNo-Dev"
@@ -20,12 +18,10 @@ export SANITY_AUTH_TOKEN="$(getSecret sanity-api-export-auth-token)"
 SAS_TOKEN="$(getSasToken $STORAGE_ACCOUNT $SUBSCRIPTION)"
 DATETIME=$(date "+%Y-%m-%d_%H%M")
 BACKUP_FILE_DATE_NAME="$DATETIME-production.tar.gz"
-BACKUP_FILE_NAME="../cms/backupfolder/production.tar.gz"
 
-RM -f "$BACKUP_FILE_NAME"
-yarn workspace cms sanity dataset export production backupfolder
+yarn workspace cms sanity dataset export production backupfolder/$BACKUP_FILE_DATE_NAME
 
 #Terraform?
 #azcopy make "https://alvnobackup.blob.core.windows.net/alvnobackup?$SAS_TOKEN"
 
-azcopy copy "$BACKUP_FILE_NAME" "https://alvnobackup.blob.core.windows.net/alvnobackup/$BACKUP_FILE_DATE_NAME?$SAS_TOKEN"
+azcopy copy "../cms/backupfolder/$BACKUP_FILE_DATE_NAME" "https://alvnobackup.blob.core.windows.net/alvnobackup/$BACKUP_FILE_DATE_NAME?$SAS_TOKEN"
--- a/packages/scripts/backup.sh
+++ b/packages/scripts/backup.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -e
+
+KEY_VAULT="alvnoprodkv"
+
+function getSecret() {
+  az keyvault secret show --vault-name $KEY_VAULT --name $1 | jq '.value' -r
+}
+
+export SANITY_AUTH_TOKEN="$(getSecret sanity-api-export-auth-token)"
+
+yarn
+yarn workspace cms sanity dataset export production backupfolder
+
+DATETIME=$(date "+%Y-%m-%d_%H%M")
+BAKCUP_FILE_NAME="production.tar.gz"
+BAKCUP_FILE_DATE_NAME="$DATETIME-production.tar.gz"
+
+echo "$BAKCUP_FILE_DATE_NAME"
+# azcopy copy "backupfolder/$BAKCUP_FILE_NAME" "https://mystorageaccount.blob.core.windows.net/mycontainer/$BAKCUP_FILE_DATE_NAME"
--- a/packages/scripts/backup.sh
+++ b/packages/scripts/backup.sh
@@ -3,19 +3,29 @@
 set -e
 
 KEY_VAULT="alvnoprodkv"
+SUBSCRIPTION="AlvNo-Dev"
+STORAGE_ACCOUNT="alvnobackup"
+SAS_EXPIRY=$(date -u -v "+1H" +"%Y-%m-%dT%H:%M:%SZ")
 
 function getSecret() {
   az keyvault secret show --vault-name $KEY_VAULT --name $1 | jq '.value' -r
 }
 
+function getSasToken() {
+  az storage account generate-sas --permissions cw --account-name $1 --services b --resource-types sco --expiry $SAS_EXPIRY --subscription $2 | jq -r
+}
+
 export SANITY_AUTH_TOKEN="$(getSecret sanity-api-export-auth-token)"
 
-yarn
+SAS_TOKEN="$(getSasToken $STORAGE_ACCOUNT $SUBSCRIPTION)"
+DATETIME=$(date "+%Y-%m-%d_%H%M")
+BACKUP_FILE_DATE_NAME="$DATETIME-production.tar.gz"
+BACKUP_FILE_NAME="../cms/backupfolder/production.tar.gz"
+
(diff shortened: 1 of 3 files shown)


packages/scripts/backup.sh
```diff
@@ -0,0 +1,22 @@
...
+
+function getSecret() {
+  az keyvault secret show --vault-name $KEY_VAULT --name $1 | jq '.value' -r
+}
+
+export SANITY_AUTH_TOKEN="$(getSecret sanity-api-export-auth-token)"
+
+yarn
+cd packages/cms
+yarn sanity dataset export production backupfolder
```
TheKnarf, line 15: Instead of `cd` I think you can run `yarn workspace cms sanity dataset export production backupfolder`

packages/scripts/backup.sh
```diff
@@ -0,0 +1,22 @@
...
+
+DATETIME=$(date "+%Y-%m-%d_%H%M")
+BAKCUP_FILE_NAME="production.tar.gz"
+BAKCUP_FILE_DATE_NAME="$DATETIME-production.tar.gz"
+
+echo "$BAKCUP_FILE_DATE_NAME"
+# azcopy copy "backupfolder/$BAKCUP_FILE_NAME" "https://mystorageaccount.blob.core.windows.net/mycontainer/$BAKCUP_FILE_DATE_NAME"
```
TheKnarf, line 22: Why is this line commentet out

TheKnarf, line 22: `backupfolder` should probably be gitignored so nobody accidentially commits the backup

packages/scripts/backup.sh
```diff
@@ -0,0 +1,22 @@
...
+
+set -e
+
+KEY_VAULT="alvnoprodkv"
+
+function getSecret() {
+  az keyvault secret show --vault-name $KEY_VAULT --name $1 | jq '.value' -r
+}
+
+export SANITY_AUTH_TOKEN="$(getSecret sanity-api-export-auth-token)"
```
TheKnarf, line 11: I'm not sure that you need to use `export` here if it's just used in the same script?

packages/scripts/backup.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -e
+
+KEY_VAULT="alvnoprodkv"
+
+function getSecret() {
```
TheKnarf, line 7: We're only using this function once, maybe combine it with the line that calls it into one line?

packages/scripts/backup.sh
```diff
@@ -0,0 +1,22 @@
...
+  az keyvault secret show --vault-name $KEY_VAULT --name $1 | jq '.value' -r
```
TheKnarf, line 8: If you run `az keyvault` with `--value` you may not need the `jq` command. That may make it easier to run in a pipeline.

TheKnarf, line 8: Add `keyvault` setup to Terraform?

packages/scripts/backup.sh
TheKnarf, line 22: Add blobstorage to Terraform?

packages/scripts/backup.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -e
```
TheKnarf, line 3: `set -e` can be combined into the shebang as `#!/usr/bin/env -S bash -e`.
- ref https://stackoverflow.com/a/57049647/359825

packages/scripts/backup.sh
TheKnarf, line 15: Should the dataset name `production` be a variable? That way we could more easily backup multiple datasets in the future.

packages/scripts/backup.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
TheKnarf, line 1: Since this is backuping `cms` maybe it should live in the `cms` folder?

packages/scripts/backup.sh
TheKnarf, line 15: Maybe replace `backupfolder` with `backupfoder/$BACKUP_FILE_NAME`?

TheKnarf, line 15: Do we need a line like `mkdir -p backupfolder`?

Trulsaa: Den var ikke dum nei. Fikser

packages/scripts/backup.sh
Trulsaa: That line is one way to go for uploading. Right now it does not work

Trulsaa: Yes. Done

packages/scripts/backup.sh
Trulsaa: Disagree. It is very likely that we will use it to get more secrets.

packages/scripts/backup.sh
Trulsaa: Yes. But not the terraform in this repo. I think it is better to put it in the infrastructure repo. When that repo is up and running.

packages/scripts/backup.sh
Trulsaa: Yes. But not the terraform in this repo. I think it is better to put it in the infrastructure repo. When that repo is up and running.

packages/scripts/backup.sh
Trulsaa: Be my guest :)

packages/scripts/backup.sh
Trulsaa: Or keep it in the scripts folder for more scripts and easyer refactoring. But be my guest.

packages/scripts/backup.sh
Trulsaa: `sanity dataset export` creates a folder not a file.

Trulsaa: Be my guest 👍

packages/scripts/backup.sh
Trulsaa: az keyvault does not recognize the --value argument

packages/scripts/backup.sh
Trulsaa: It was tried both ways. It only worked with export

packages/scripts/backup.sh
TheKnarf: I do not want a generic script folder with a bunch of unrelated scripts.

Trulsaa: I was thinking related ones

packages/scripts/backup.sh
TheKnarf: As the infrastructure folder is not yet up, for now; all alv-website infrastructure should go into this repo

packages/scripts/backup.sh
TheKnarf: Hm, thats not what it looks like when I run the help command `yarn workspace cms sanity dataset export --help`:

```
usage: sanity dataset export [NAME] [DESTINATION]

   Export dataset to local filesystem as a gzipped tarball

Options
  --raw                     Extract only documents, without rewriting asset references
  --no-assets               Export only non-asset documents and remove references to image assets
  --no-drafts               Export only published versions of documents
  --no-compress             Skips compressing tarball entries (still generates a gzip file)
  --types                   Defines which document types to export
  --overwrite               Overwrite any file with the same name
  --asset-concurrency <num> Concurrent number of asset downloads

Examples
  sanity dataset export moviedb localPath.tar.gz
  sanity dataset export moviedb assetless.tar.gz --no-assets
  sanity dataset export staging staging.tar.gz --raw
  sanity dataset export staging staging.tar.gz --types products,shops
```

The examples seems to imply that you should give it a filename.

packages/scripts/backup.sh
TheKnarf: Then the folder should not be named `scripts`, but rather what makes them related. In this case I want it in the `cms` folder.

packages/scripts/backup.sh
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export SANITY_AUTH_TOKEN="$(getSecret sanity-api-export-auth-token)"
+
+SAS_TOKEN="$(getSasToken $STORAGE_ACCOUNT $SUBSCRIPTION)"
+DATETIME=$(date "+%Y-%m-%d_%H%M")
+BACKUP_FILE_DATE_NAME="$DATETIME-production.tar.gz"
+BACKUP_FILE_NAME="../cms/backupfolder/production.tar.gz"
+
+RM -f "$BACKUP_FILE_NAME"
```
TheKnarf, line 25: Let's not delete the file. How about we just rename backups before uploading them to the name that have DATETIME in it?

TheKnarf (approved):

PR #402, opened 2021-05-21
Playlist seasons

mikkelcodes:


packages/website/gatsby-node.js
```diff
@@ -1,203 +1,47 @@
...
-}
+const videoTemplate = path.resolve(`./src/templates/video.js`);
+const careerTemplate = path.resolve(`./src/templates/career.js`);
+const articleTemplate = path.resolve(`./src/templates/article.js`);
+const serviceTemplate = path.resolve(`./src/templates/service.js`);
+const categoryTemplate = path.resolve(`./src/templates/category.js`);
+const videoCategoryTemplate = path.resolve(`./src/templates/videoCategory.js`);
+const videoserieTemplate = path.resolve(`./src/templates/videoserie.js`);
 
+let videoseries;
```
TheKnarf, line 13: This probably doesn't need to be defined outside of `createPages`, it can also probably be made const.

packages/website/gatsby-node.js
```diff
@@ -239,54 +83,90 @@ exports.createPages = async ({ graphql, actions }) => {
...
+                lower: true,
+              });
+              video.formattedPublishedAt = new Date(
+                video.publishedAt
+              ).toUTCString();
+              return video;
+            })
+            .sort((a, b) => (a.position > b.position ? -1 : 1));
+
+          seasons.push(seasonVideos);
```
TheKnarf, line 133: Both pushing it to an array and returning it seems like a mistake. It's probably enough to just return it?

packages/website/gatsby-node.js
```diff
@@ -239,54 +83,90 @@ exports.createPages = async ({ graphql, actions }) => {
...
+          return seasonVideos;
+        })
+      ).then(() => {
```
TheKnarf, line 136: Mixing `.then` and `await` is bad practice. Use one or the other.

TheKnarf (changes requested):
Good work getting this working! 🎉

The build is not building on the buildserver, so let's try and fix that. And also let's spend a bit time to clean up some of the code. But otherwise great job 😀

I'm looking forward to getting this merged in 🚢

packages/website/src/templates/video.js
```diff
@@ -4,20 +4,14 @@ import Layout from '../layout';
 import { window } from 'browser-monads';
 import { SocialShare } from '../components/socialShare';
 import { VideoEpisode } from '../components/videoEpisode';
-import { useVideoseriesQuery } from '../hooks/useVideoseriesQuery';
 
 const VideoTemplate = ({ pageContext }) => {
-  const { video, list, playlistName, playlistSlug } = pageContext;
+  const { video, season, playlistName } = pageContext;
   const [playlist, setPlaylist] = useState(null);
```
TheKnarf, line 10: This does'nt really seem to be in use, its just caling `setPlaylist(season)`. Just remove the whole `useState` and `useEffect` and rename the rest of the code to use the `season` variable directly.

packages/website/src/templates/videoCategory.js
```diff
@@ -1,45 +1,41 @@
...
-import { EpisodeCards } from '../components/episodeCards';
 import { VideoHero } from '../components/videoHero';
-import slugify from 'slugify';
+import { VideoFilter } from '../components/videoFilter';
+import { VideoSection } from '../components/videoSection';
 
-const VideoCategoryTemplate = ({ pageContext, data }) => {
-  const { title, videos, slug } = pageContext;
+const VideoCategoryTemplate = ({ pageContext }) => {
   const [playlist, setPlaylist] = useState(null);
```
TheKnarf, line 9: Again, this `useState` and `useEfffect` can probably be removed. As it's basicly just setting the `playlist` variable to the content of the `seasons` variable... Just rename the rest of the code to use `seasons` and remove it.

packages/website/src/templates/videoserie.js
```diff
@@ -0,0 +1,72 @@
...
+import { VideoFilter } from '../components/videoFilter';
+
+const Videoseries = ({ pageContext }) => {
+  const { videoseries } = pageContext;
+  const data = useVideoseriesQuery();
+
+  const {
+    sanityVideoseriesPage: { pageDescription } = { pageDescription: false },
+    sanityVideoseriesPage: { pageTitle } = { pageTitle: false },
+  } = data;
```
TheKnarf, line 16: Use destructuring here just makes the code a lot harder to read. I struggle to understand what these 4 lines of code does. If I understand it correctly then writing it as following would be alot easier to read:

```
const pageDescription = data.sanityVideoSeriesPage.pageDescription || false;
const pageTitle = data.sanityVideoSeriesPage.pageTitle || false;
```

packages/website/src/templates/videoserie.js
```diff
@@ -0,0 +1,72 @@
...
+          ))}
+        </div>
+      </div>
+    </Layout>
+  );
+};
+
+export default Videoseries;
+
+const Videoserie = ({ videoserie, data, index }) => {
```
TheKnarf, line 44: If the `Videoseries` component is the default export then it should probably be on the bottom of the file and this component on the top.

packages/website/src/templates/videoserie.js
```diff
@@ -0,0 +1,72 @@
...
+  const [sortedVideos, setSortedVideos] = useState(videoserie.seasons[0]);
+  const sortedList = (list) => {
+    setSortedVideos(list);
+  };
+  return (
+    <div key={`videoserie-${index}`}>
+      <>
```
TheKnarf, line 51: The `<>` is unnecessary since its inside a `<div>`

packages/website/src/components/videoFilter/index.js
```diff
@@ -48,34 +50,18 @@ export const VideoFilter = ({ videos, tabs, onChange }) => {
...
-      id="3"
-    >
-      Sesong 3
-    </div>
+    {seasons.map((season, i) => (
+      <div
+        className={`font-semibold mr-6 pb-1 border-b-2  ${
+          activeSeasonIndex === i ? 'border-yellow' : 'border-navy'
+        } cursor-pointer`}
+        onClick={onClick}
```
TheKnarf, line 60: `div`s should not have `onClick` events. Thats not in accordance with accessibility. It needs to be either a `<button>` or an `<a>`.

packages/website/src/components/videoFilter/index.js
```diff
@@ -1,19 +1,17 @@
...
 
   const updateVideos = (updatedVideos) => {
-    setActive(updatedVideos);
     onChange(updatedVideos);
   };
 
   useEffect(() => {
     const sortedActive = [];
-    active.forEach((el) => sortedActive.push(el));
+    seasons[activeSeasonIndex].forEach((el) => sortedActive.push(el));
```
TheKnarf, line 14: Instead of:

```
const sortedActive = [];
seasons[activeSeasonIndex].forEach((el) => sortedActive.push(el));
```

Use `map` as following:

```
const sortedActive = seasons[activeSeasonIndex].map(el => el);
```

or simplify it further to just:

```
const sortedActive = [ ...seasons[activeSeasonIndex] ];
```

packages/website/src/components/videoFilter/index.js
```diff
@@ -1,19 +1,17 @@
...
 import { SortField } from '../blogFilter';
 
-export const VideoFilter = ({ videos, tabs, onChange }) => {
+export const VideoFilter = ({ seasons, onChange }) => {
   const [sort, setSort] = useState(null);
-  const [active, setActive] = useState(videos);
-  const [season, setSeason] = useState('1');
+  const [activeSeasonIndex, setActiveSeasonIndex] = useState(0);
 
   const updateVideos = (updatedVideos) => {
```
TheKnarf, line 8: `updateVideoes` just calls `onChange` now so the whole function can just be deleted and replaces with calling `onChange` directly.

TheKnarf (approved):

PR #413, opened 2021-06-09
Update phone link from org.no to phone

timorzadir:
The phone link tries to open a tel-link to the organisation number. If that's not something new I missed, I think we should link to the phone number.

### DIFF
--- a/packages/website/src/pages/kontakt-oss.js
+++ b/packages/website/src/pages/kontakt-oss.js
@@ -65,7 +65,7 @@ const Contact = ({ location }) => {
               <div className="hidden xl:block">
                 <div className="text-lg tracking-wider 2xl:ml-30 mb-20">
                   <div className="mb-3">
-                    <a href="tel:822704042" className="flex">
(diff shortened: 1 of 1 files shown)


packages/website/src/pages/kontakt-oss.js
```diff
@@ -65,7 +65,7 @@ const Contact = ({ location }) => {
               <div className="hidden xl:block">
                 <div className="text-lg tracking-wider 2xl:ml-30 mb-20">
                   <div className="mb-3">
-                    <a href="tel:822704042" className="flex">
+                    <a href="tel:91530363" className="flex">
```
TheKnarf, line 68: I think we have actually added the phone number to Sanity now? So instead of changing it from one hardcoded one to another we could change it to use the one from Sanity CMS.

TheKnarf (approved):

packages/website/src/pages/kontakt-oss.js
timorzadir: Good idea, I created issue #414 for it.

PR #421, opened 2021-06-15
Update gatsby

mikkelcodes:
🛳


package.json
```diff
@@ -31,5 +31,8 @@
       "pre-commit": "run-s -s lint"
     }
   },
-  "prettier": "@blueberrypie/prettier-config"
+  "prettier": "@blueberrypie/prettier-config",
+  "dependencies": {
+    "gatsby-plugin-image": "^1.7.1"
```
TheKnarf, line 36: Is it correct to add this to the root-`package.json` or should it be inn `packages/website/package.json`?

packages/cms/schemas/landingPage.js
```diff
@@ -40,5 +40,88 @@ export default {
       title: 'Video Introduction - mp4',
       type: 'file',
     },
+    {
```
TheKnarf, line 43: Why did all of these schema changes come in a PR thats supposed to be about updating Gatsby?

package.json
mikkelcodes: packages/website/package.json is correct. I'll update

packages/cms/schemas/landingPage.js
mikkelcodes: I got unreliable results when querying for single images after the v3 update for some reason. Sanity sourced images, on the other hand, works perfectly. So the quickest way forward for me was to source these landing page images from Sanity instead. I believe sourcing these images from Sanity is the ideal configuration we want anyway?

But I agree that it could have been in a separate PR and linked to its own issue

TheKnarf (approved):

PR #460, opened 2021-09-08
Rewrite dev pipeline to use the new run script to deploy website

Trulsaa:
Depends on #430

### DIFF
--- a/.AzureDevOps/pipelines/azure-pipelines-dev.yml
+++ b/.AzureDevOps/pipelines/azure-pipelines-dev.yml
@@ -6,23 +6,31 @@ pr: none
 pool:
   vmImage: ubuntu-latest
 
-variables:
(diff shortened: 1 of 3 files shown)


TheKnarf:
Does this depend on #430?

timorzadir (approved):

TheKnarf (changes requested):
I'm blocking this PR until #430 is merged.

TheKnarf:
If we're going over to Kubernetes then I have two questions:

- Could we at first only change the `dev` pipeline so that we get to test this for a while?
- Could we replace Azure DevOps and only use Github Actions?

Trulsaa:
Only changing the `dev` pipeline is a good idea 👍  Replacing Azure DevOps with Github Actions is not something that is needed now and will only delay this task unnecessarily. But since we are now toing to a bash script it should be easier to do it as a task further down the line.

TheKnarf (approved):

PR #492, opened 2021-09-14
Initialise shared component structure. Closes #491

### LINKED ISSUE #491: Create shared component library

mikkelcodes:


packages/shared-components/videoIntro/index.js
```diff
@@ -1,10 +1,15 @@
 import React from 'react';
```
TheKnarf, line 1: If we create a `shared-components` folder in `packages/` then we should probably first create a PR that sets this folder up. Run `yarn init -y` in it to get a package.json, create a `src/`-folder and a `Readme.md` file.

TheKnarf (changes requested):

PR #494, opened 2021-09-14
About component

mikkelcodes:


packages/shared-components/videoIntro/index.js
```diff
@@ -1,10 +1,15 @@
...
+export const VideoIntro = ({
+  videoMp4,
+  videoWebm,
+  children,
+  textColor = 'white',
+}) => {
   return (
     <div
-      className="lg:pr-12 lg:grid mx-auto max-w-1440 sm:pt-15"
+      className={`lg:pr-12 lg:grid mx-auto max-w-1440 sm:pt-15 text-${textColor}`}
```
TheKnarf, line 12: Changes like text-color could be added in its own PR, before we start moving components.

PR #502, opened 2021-09-17
add alv-b frontend files

mikkelcodes:


packages/website-alvb/gatsby-config.js
```diff
@@ -0,0 +1,67 @@
...
+    siteUrl: `https://www.alvb.no`,
+    title: `Eksperimentell kreftbehandling for hund & katt`,
+    description: `Vi ønsker å skape et globalt ‘center of excellence’ gjennom vår forskning på eksperimentell kreftbehandling for hund & katt.`,
+    author: `Alv`,
+  },
+  plugins: [
+    {
+      resolve: `gatsby-plugin-google-analytics`,
+      options: {
+        trackingId: 'UA-197345464-1',
```
TheKnarf, line 16: alv-b might want their own google analytics

packages/website-alvb/gatsby-config.js
```diff
@@ -0,0 +1,67 @@
...
+    'gatsby-plugin-image',
+    'gatsby-transformer-sharp',
+    'gatsby-plugin-sharp',
+    'gatsby-plugin-postcss',
+    'gatsby-plugin-react-helmet',
+    `gatsby-plugin-sitemap`,
+    {
+      resolve: `gatsby-source-sanity`,
+      options: {
+        projectId: `mnr37rl0`,
```
TheKnarf, line 29: while this might be fine for testing, we'll need to change this for production

packages/website-alvb/src/hooks/useWindowDimensions.js
```diff
@@ -0,0 +1,27 @@
+import { useState, useEffect } from 'react';
```
TheKnarf, line 1: Is there any way we could rewrite whatever code uses this hook so that this type of logic isn't handled in JavaScript but in CSS? I would very much like for us to delete this hook.

packages/website-alvb/src/templates/videoserie.js
```diff
@@ -0,0 +1,89 @@
+import React, { useState } from 'react';
```
TheKnarf, line 1: Do we have `videoserie` on alv-b?

packages/website-alvb/src/templates/videoCategory.js
```diff
@@ -0,0 +1,76 @@
+import React, { useState } from 'react';
```
TheKnarf, line 1: Do we have `videoCategory` on alv-b?

packages/website-alvb/src/templates/video.js
```diff
@@ -0,0 +1,87 @@
+import React from 'react';
```
TheKnarf, line 1: Do we have `video` on alv-b?

packages/website-alvb/gatsby-config.js
mikkelcodes: Yes, this is actually their own analytics id

packages/website-alvb/src/hooks/useWindowDimensions.js
mikkelcodes: I have replaced useWindowDimensions with css in 3 out of 4 components. The remaining component is gridContainer (shared). I propose this becomes a separate issue to refactor this last component as it will take more time. I'll have to rewrite the logic of how the employee cards / bio section works. What do you say?

TheKnarf: Creating a seperate task for it sounds good.

packages/website-alvb/gatsby-config.js
mikkelcodes: I created a new project under the organisation you created. I sent you an invite to it

packages/website-alvb/src/templates/videoserie.js
mikkelcodes: I removed this template file

packages/website-alvb/src/templates/video.js
mikkelcodes: I removed this template file

TheKnarf:
Do we need all of the image files on alv-b? Seems like some of them are only relevant for alv.

packages/website-alvb/src/hooks/useWindowDimensions.js
TheKnarf: If gridContainer is shared then maybe we can delete useWindowDimensions from `website-alvb`?

TheKnarf:
Still seems like theres images we don't need in alv-b, perhaps we could use some kind of placeholder image service if all we're using the images for is placeholders while developing? Something like https://placekitten.com/

mikkelcodes:
> Still seems like theres images we don't need in alv-b, perhaps we could use some kind of placeholder image service if all we're using the images for is placeholders while developing? Something like https://placekitten.com/

Yes, kept them for the sake of having some placeholders but we can use placekitten instead

TheKnarf:
> packages/website-alvb/src/assets/ALVB-negative-logo.png

Maybe this file should be an svg? And it should probably live in `packages/website-alvb/src/components/logo/`.

packages/website-alvb/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "react-dom": "^17.0.1",
+    "react-headroom": "^3.1.0",
+    "react-helmet": "^6.1.0",
+    "react-reveal": "^1.2.2",
+    "react-share": "^4.3.1",
+    "react-slick": "^0.28.0",
+    "react-three-fiber": "^5.3.17",
+    "slugify": "^1.4.7",
+    "tailwindcss": "latest",
+    "three": "^0.124.0",
```
TheKnarf, line 38: I belive `three` is not used in alv-b.

packages/website-alvb/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "postcss": "latest",
+    "react": "^17.0.1",
+    "react-device-detect": "^1.15.0",
+    "react-dom": "^17.0.1",
+    "react-headroom": "^3.1.0",
+    "react-helmet": "^6.1.0",
+    "react-reveal": "^1.2.2",
+    "react-share": "^4.3.1",
+    "react-slick": "^0.28.0",
+    "react-three-fiber": "^5.3.17",
```
TheKnarf, line 35: I belive `react-three-fiber` is not used in alv-b.

packages/website-alvb/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "build": "gatsby build",
+    "serve": "gatsby serve",
+    "dev": "gatsby develop",
+    "clean": "gatsby clean",
+    "disable-telemetry": "gatsby telemetry --disable",
+    "start:storybook": "start-storybook -p 6006 --no-dll",
+    "build:storybook": "NODE_ENV=test build-storybook --no-dll --quiet",
+    "test": "jest"
+  },
+  "dependencies": {
```
TheKnarf, line 13: Maybe we should go through and strip out dependencies thats not needed for alv-b?

packages/shared-components/src/components/logo/index.jsx
```diff
@@ -1,10 +1,12 @@
 import React from 'react';
```
TheKnarf, line 1: Maybe Logo shouldn't be a shared component? Since conceptually Alv and Alv-b don't share the same Logo.

mikkelcodes: Yes, clearly

TheKnarf (approved):

PR #515, opened 2021-09-22
Cms alvb

mikkelcodes:


packages/cms-alvb/Readme.md
```diff
@@ -0,0 +1 @@
+# Sanity CMS
```
TheKnarf, line 1: Maybe update description to identify that it have something to do with alv-b? Also add the link in the root Readme like we have done for the other packages.

mikkelcodes: I added references to alv b in readme files

TheKnarf (approved):

PR #577, opened 2021-11-02
revert slug change from /om-oss to /selskapet

mikkelcodes:


packages/shared-components/src/components/aboutIntro/index.js
```diff
@@ -59,10 +67,10 @@ export const AboutIntro = ({ topImg, bottomImg }) => {
             </h3>
             <div className="w-14 h-3px bg-yellow mt-6px mb-4" />
             <div className="divide-y-2">
-              <ReadMoreLink link="/selskapet/livet-i-alv">
+              <ReadMoreLink link="/om-oss/livet-i-alv">
```
TheKnarf, line 70: Is it a problem that this link is hard-coded in a shared component?

TheKnarf (approved):

packages/shared-components/src/components/aboutIntro/index.js
mikkelcodes: This component is not supposed to be in shared-components folder. It is only used by alv.no. I'll create an issue to update that.

PR #588, opened 2021-11-12
Fix cta image rendering bug

mikkelcodes:


packages/shared-components/src/components/richTextTypes/index.js
```diff
@@ -13,14 +13,10 @@ const urlFor = (source) =>
 
 const Image = ({ props, link }) => (
   <a href={link} className={styles.wrapper}>
-    <div
+    <img
       alt={props.node.heading}
       className={styles.img}
-      style={{
-        backgroundImage: `url(${urlFor(props.node.image.asset)
```
TheKnarf, line 20: Consider moving the url out in its own variable to make the code easier to read.

TheKnarf (approved):

PR #615, opened 2021-12-12
Add switch locale button and updates according to new design spec

mikkelcodes:


packages/website-alvb/src/components/header/index.js
```diff
@@ -24,8 +27,43 @@ export const Header = ({
     setOpen(false);
   };
 
-  const crumbs = window.location.pathname.split('/');
-  crumbs.shift();
+  const pathStr = window.location.pathname.replace('%C3%A5', 'å');
```
TheKnarf, line 30: If we have this problem with `å`, then I'd imagine we got the same problem with `æ` and `ø`.

TheKnarf (approved):

PR #622, opened 2022-01-12
add localeCta to hero

mikkelcodes:


packages/cms-alvb/schemas/heroCta.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
```
TheKnarf, line 1: Why is this in the `cms-alvb` package?

mikkelcodes: Good Q

TheKnarf (approved):

PR #626, opened 2022-02-03
Added google-tag-manager to ALV-B

klyngen:
Added gatsby-plugin-google-tagmanager
- Verified that the script-tag is added to the public-content after the build


Trulsaa:
Jeg fant en feil. Se om du finner den du også.

Trulsaa (approved):
LGTM

PR #632, opened 2022-02-08
Server side rendering for Blogg page and all articles

klyngen:
# Server Side Rendering

* Steinar has implemented a varnish cache to ensure performance
* Blogg-pages has been converted into using SSR

This is the first step to converting the whole page


TheKnarf:
Why is there an empty file [packages/website/.projectile](https://github.com/Alv-no/alv-website/pull/632/files#diff-6a113336f4a25530f82724c3268171d11abd30d31546d58128669aae16825f5f) here?

website.Dockerfile
```diff
@@ -1,6 +1,10 @@
-# Stage 1 - Building the website
-FROM node:14
+FROM nginx:1.20.2-alpine
 
+# Stage 1 - Install node, yarn and python
```
TheKnarf, line 3: Maybe don't call it `stage 1` -> `stage 2` etc here. Stages in Docker refer to multiple uses of the `FROM` statement. If you only have 1 `FROM` statement then you don't have a mulitistage Docker build. If you do have two `FROM` statements then anything starting from the first `FROM` statement is the first stage of a multistage Dockerfile, while everything starting from the second `FROM` is the second stage, etc.

website.Dockerfile
```diff
@@ -24,9 +28,17 @@ ADD "https://www.random.org/cgi-bin/randbyte?nbytes=10&format=h" skipcache
...
-FROM nginx:1.19-alpine
-COPY --from=0 /app/packages/website/public /usr/share/nginx/html/
+# Stage 3 - Setup nginx webserver
+RUN cp -r /app/packages/website/public/* /usr/share/nginx/html/
 COPY website.nginx.conf /etc/nginx/conf.d/default.conf
-
 EXPOSE 80
+
+# Stage 4 - Setup and install Varnish
+RUN apk add --update varnish
```
TheKnarf, line 37: Probably better to gather all of the `apk` steps for installing stuff at the top of the file. Docker caches each step in a Dockerfile as its own layer, so gathering intesive steps that seldomly changes in the top is a good practice. That way when the source code of our project changes and the Docker container needs to be rebuild it doesn't need to redo these steps since it can see that their cached.

If we do a `apk add` after the build step for our website then it can't cache this step and need to redo it everytime we build. But if we put it at the top then it won't need to redo it.

website.Dockerfile
```diff
@@ -24,9 +28,17 @@ ADD "https://www.random.org/cgi-bin/randbyte?nbytes=10&format=h" skipcache
...
+COPY default.vcl /etc/varnish/
+
+# Stage 5 - Setup and run supervisor
+RUN apk update && apk add --no-cache supervisor
```
TheKnarf, line 41: Move up to top. Gather `apk update` and `apk add` togheter. See other comment.

website.Dockerfile
```diff
@@ -1,6 +1,10 @@
-# Stage 1 - Building the website
-FROM node:14
+FROM nginx:1.20.2-alpine
 
+# Stage 1 - Install node, yarn and python
+RUN apk add --update nodejs yarn
+RUN apk add --no-cache python3 make gcc g++
```
TheKnarf, line 5: You could consider merging these two `apk add` steps together into a single `RUN` statement. That way Docker will only have to create one cache layer for the `apk`-install step instead of multiple.

website.Dockerfile
klyngen: Good point. This dockerfile could probably also be a multistage build. However, that is work for another time.

website.Dockerfile
klyngen: Agreed. Bundled the package-manager-commands together. Many has become one.

website.Dockerfile
klyngen: Should be in the oneliner

klyngen:
> Why is there an empty file [packages/website/.projectile](https://github.com/Alv-no/alv-website/pull/632/files#diff-6a113336f4a25530f82724c3268171d11abd30d31546d58128669aae16825f5f) here?

Emacs-users (me) defines projects by either having a git-repo or folders with an .projectile-file. Those files should not be committed. I like dividing the project into nice small sub-projects.

PR #634, opened 2022-02-15
Added apollo-client queries for blogg and article

klyngen:
Page queries are only ran once during build time. In order to get true server-side rendering, we need to load data using the `getServerData`-method. That also means that we need to do queries using apollo-client


packages/website/src/hooks/useBlogQueryServerSide.js
```diff
@@ -0,0 +1,72 @@
+import { gql } from '@apollo/client';
+import { createGatsbyImages } from '../server-side/imageCreator';
+import { client } from '../server-side/client';
+export async function getBlogDataServerSide() {
```
TheKnarf, line 4: Presumably these aren't reused by different pages so perhaps they can live in the page where their used?

packages/website/src/server-side/client.js
```diff
@@ -0,0 +1,9 @@
+import { ApolloClient, InMemoryCache } from '@apollo/client';
+
+export const client = new ApolloClient({
+  uri: `https://mnr37rl0.apicdn.sanity.io/v1/graphql/production/default`,
```
TheKnarf, line 4: Maybe we also can stick the first part of the url, the part thats hardcoded to sanity dataset in some variable? If not, then we're making it harder and harder to eventually add a switch so that we can choose between using the `development` and `production` dataset.

packages/website/src/server-side/imageCreator.js
```diff
@@ -0,0 +1,54 @@
...
+  let assets = {};
+
+  if (asset.metadata?.dimensions) {
+    assets = {
+      ...asset.metadata.dimensions,
+    };
+  }
+
+  return getGatsbyImageData(node, assets, {
+    projectId: 'mnr37rl0',
```
TheKnarf, line 26: Again; dataset and projectId, can we have the in a single place in the code / env variable?

packages/website/src/hooks/useBlogQueryServerSide.js
klyngen: Blog query is used by both `pages/blogg.js` and `teamplate/articje.js` but your point is true for the articleQuery.

packages/website/src/server-side/client.js
klyngen: Agreed. I projectId and dataset as environmentVariables :+1:

klyngen:
It does not build due to the lack of parameters in azure keyvault

run
```diff
@@ -13,8 +13,7 @@ CONTAINER_REGISTRY_NAME=alvkubernetesclustertestacr
 CONTAINER_REGISTRY="$CONTAINER_REGISTRY_NAME.azurecr.io"
 SERVICE_TO_BUILD="$3"
 SUBSCRIPTION="k8s-$ENV-subscription"
-
-function getSecret() {
+function getSe"bcret() {
```
TheKnarf, line 16: Typo?

run
```diff
@@ -146,6 +146,6 @@ if [ "deploy" = "$1" ]; then
   getSecrets
   az acr login --name $CONTAINER_REGISTRY_NAME
   buildServices
-  pushServices
-  deployServices
+  #pushServices
+  #deployServices
```
TheKnarf, line 150: Why do we no longer push and deploy services?

run
klyngen: This is the result of two developers sitting up late and trying to edit code. And then typing on the keyboard like a monkey

PR #638, opened 2022-03-03
Jobbe i alv sider med SSR

klyngen:
Karriere-sider


run
```diff
@@ -13,7 +13,7 @@ CONTAINER_REGISTRY_NAME=alvkubernetesclustertestacr
 CONTAINER_REGISTRY="$CONTAINER_REGISTRY_NAME.azurecr.io"
 SERVICE_TO_BUILD="$3"
 SUBSCRIPTION="k8s-$ENV-subscription"
-function getSe"bcret() {
```
TheKnarf, line 16: I don't understand this diff.

klyngen: I agree. I removed it from the commit.

mikkelcodes (approved):
As we've discussed, we're moving in the direction of using an SSG framework to create an SSR website, which doesn't make all that much sense :D But, I say we merge this as a temporary solution to the issue at hand and seek to either convert over to NextJS/Remix or implement Sanity Live Preview in combination with a build-by-webhook solution (which according to Jakob is a perfectly ok solution to the same problem).

Looks good 💯

PR #709, opened 2022-06-21
Sitemap

mikkelklokkerud:
- Issue: No sitemap entries for article pages - these are now generated
- Issue: Unwanted sitemap entry containing "[slug]" in its url - removed
- Issue: All sitemap entries previously had the same priority value (0.7) - they now use values as chosen by marketing

### DIFF
--- a/packages/website/src/pages/blogg.js
+++ b/packages/website/src/pages/blogg.js
@@ -32,7 +32,7 @@ const Blog = ({ data, serverData }) => {
   return (
     <Layout
       whiteIcons
-      pageTitle="Blogg"
+      pageTitle={serverData.articles.pageMetadata.pageTitle}
       pageDescription={serverData.articles.pageMetadata.pageDescription}
       layoutData={layoutData}
     >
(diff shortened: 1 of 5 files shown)


packages/website/src/pages/blogg.js
```diff
@@ -32,7 +32,7 @@ const Blog = ({ data, serverData }) => {
   return (
     <Layout
       whiteIcons
-      pageTitle={serverData.articles.pageMetadata.pageTitle}
+      pageTitle="Blogg"
```
klyngen, line 35: Is it possible to not hard-code this item?

mikkelklokkerud: Yup, however it was just a short term fix that was approved by jakob. I'll remove it as it's not really related to the scope of the PR

mikkelklokkerud, line 35:
```suggestion
      pageTitle={serverData.articles.pageMetadata.pageTitle}
```

klyngen (approved):

PR #768, opened 2022-10-17
add success and err msg

mikkelklokkerud:
- Callback function is simplified and returns request response
- Indicates to user if email submission was successful or not


packages/shared-components/src/components/companyRichText/utils.js
```diff
@@ -1,20 +1,18 @@
-import { window } from 'browser-monads';
```
TheKnarf, line 1: I think you still want to keep this line?

mikkelklokkerud: Yess, true

TheKnarf (approved):

PR #771, opened 2022-10-19
Fix meta info bug

mikkelklokkerud:


mikkelklokkerud:
Fixes bug where wrong meta info was showing for alvb

TheKnarf (approved):
Squash?

PR #773, opened 2022-10-19
replace video with image

mikkelklokkerud:


mikkelklokkerud:
Replace video with image as per design spec

TheKnarf (approved):

PR #787, opened 2022-11-12
Remove the Github action for deploying Storybook to Github Pages

TheKnarf:
We never kept Storybook up to date, and now that we have multiple packages and a shared-components folder, deploying to Github Pages is no longer a good solution.

If we do decide to update Storybook and keep using it then I would recomend to instead deploy it directly to Kubernetes with the rest of our apps.


TheKnarf:
We can also delete the `gh-pages` branch after this is merged.

klyngen (approved):

PR #793, opened 2022-11-21
Hire text

mikkelklokkerud:


packages/shared-components/src/components/container/index.js
```diff
@@ -41,9 +41,7 @@ export const Container = ({
         !removePaddingTop && topSpacing
       } ${!removePaddingBottom && bottomSpacing} px-5 sm:px-12`}
     >
-      <div className={`max-w-${maxWidth || '1200'} mx-auto overflow-hidden`}>
-        {children}
-      </div>
+      <div className={`max-w-${maxWidth || '1200'} mx-auto`}>{children}</div>
```
klyngen, line 44: Here we have conditionals that decide the tailwind-css-classes. This might not look correct after the tailwind-pruning. I suggest using the styles-property to override `max-w-1200`

packages/shared-components/src/components/hire/index.js
```diff
@@ -127,3 +111,29 @@ export const HireAlt = ({ image, text, title, imageText, blue }) => {
...
+  return (
+    <div
+      className={`twelve:-mr-1/2 lg:flex flex-col hidden lg:-mr-48 w-full px-0 lg:-ml-10 xl:-ml-5 items-center justify-center`}
+    >
+      <h2 className="uppercase text-slider mb-8 font-semibold tracking-wider lg:w-4/6">
+        {title}
+      </h2>
+      <p
+        className={`text-footer sm:text-xl tracking-wider ${
+          buttonText && 'mb-16'
```
klyngen, line 125: Conditionals and tailwind-css-pruning is a combo that works perfectly during development and can supprise you in prod.

klyngen (changes requested):

klyngen:
On a more general note. The `Hire`-component is used by both Alv's. Personally I like how the Hire-component looks now, I guess that change is good as long as it's fine with Jakob.

packages/shared-components/src/components/hire/index.js
mikkelklokkerud: According to the same [tailwindcss docs](https://tailwindcss.com/docs/content-configuration#dynamic-class-names) as well as our currently deployed websites this works just fine. However, I'm changing from && operator to ternary to avoid `false` as a classname if `buttonText` is fasly. 

Sounds ok?

mikkelklokkerud:
> On a more general note. The `Hire`-component is used by both Alv's. Personally I like how the Hire-component looks now, I guess that change is good as long as it's fine with Jakob.

Great. And this UI change has been approved by the marketing department

packages/shared-components/src/components/container/index.js
klyngen: Sounds great :+1:

packages/shared-components/src/components/hire/index.js
klyngen: Sound good! Simplifies bughunting in the future :+1:

klyngen (approved):

PR #794, opened 2022-11-21
Remove norwegian version

mikkelklokkerud:


packages/shared-components/src/components/featuredCard/index.js
```diff
@@ -23,7 +23,7 @@ export const FeaturedCard = ({
 
   return (
     <Link
-      to={`/blogg/${slug.current}`}
+      to={`/${isEnLocale ? 'blog' : 'blogg'}/${slug.current}`}
```
klyngen, line 26: Instead of the isEnLocale-logic, would it be better to pass a post-prefix? AlvA would pass `blogg` and AlvB would pass `blog`. That way, the shared component is `multi-lingual` instead of `two-lingual`

packages/shared-components/src/components/postCard/index.js
```diff
@@ -28,7 +28,7 @@ export const PostCard = ({
 
   return (
     <Link
-      to={`/blogg/${slug.current}`}
+      to={`/${isEnLocale ? 'blog' : 'blogg'}/${slug.current}`}
```
klyngen, line 31: If you agree with the comment above. Do the same thing here

klyngen (changes requested):
Hey! Left a few comments on the PR very simple ones. I see that there is some tech-debt on the webpages after this. Probably some schemas with localization in them. I will make issues. For now this is good

packages/shared-components/src/components/featuredCard/index.js
mikkelklokkerud: Yes I agree that it's better practice to make language specific input more dynamic though it means more props drilling. I have updated for this now. A few isEnLocale remains to be defeated I suspect

I've also used something like this in the past: 
https://github.com/i18next/react-i18next/

But didn't want to include it yet to avoid unnecessary js load

klyngen: Yeah. Lets take the remaining tech-debt later :+1:

klyngen (approved):

PR #797, opened 2022-11-22
rename services to products

mikkelklokkerud:
New information: rename "services" to "products"


klyngen (approved):
In my mind this should work :+1:

PR #810, opened 2022-12-05
fix color rendering bugs

mikkelklokkerud:
Both rounded button bug and title colors were caused by the same type of error: competing classnames were added and overrides happened :(

But it's fixed :)

Also updated some conditional logic to adhere to tailwind docs ✌️


TheKnarf (approved):

TheKnarf:
Se om du bør oppdatere branchen før du merger, ligger er en liten merge commit inni der.

```
git fetch
git rebase origin/master
git push --force-with-lease
```

PR #814, opened 2022-12-06
add initial blog filter by query param

mikkelklokkerud:
Makes it possible to have urls set an initial filter config for blog articles with for example:
- alv.no/blogg/?kategori=someCategoryHere

Also fixed an issue where the article card "Read more" text was overlaying the filter background.

### DIFF
--- a/packages/shared-components/src/components/blogFilter/index.js
+++ b/packages/shared-components/src/components/blogFilter/index.js
@@ -1,7 +1,7 @@
-import React, { useState, useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
+import { FilterContainer } from '../filterContainer';
 import { DropdownMini, Tag } from '../icon';
 import * as styles from './BlogFilter.module.css';
-import { FilterContainer } from '../filterContainer';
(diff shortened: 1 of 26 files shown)


packages/shared-components/src/components/blogFilter/index.js
```diff
@@ -11,33 +11,24 @@ export const BlogFilter = ({
...
-    } else if (sort === 'newest') {
-      sortedActive.sort((a, b) => {
-        if (a.rawDate > b.rawDate) {
-          return -1;
-        } else {
-          return 1;
-        }
-      });
-    }
+    const sortedActive = [...active];
```
klyngen, line 25: This mechanism is better then what we got at the moment. Buuuut it is reversed. When I click `Newest to Oldest` I get the total oposite

mikkelklokkerud: Yes, I managed to sort to newest regardless of sorting option.. It's fixed !  🙏

klyngen (approved):

PR #819, opened 2022-12-08
Update azure-pipelines-review-killer.yml for Azure Pipelines

klyngen:


mikkelklokkerud (approved):
I'm not sure what this line does but it most likely makes the world a better place

PR #823, opened 2022-12-09
dynamic menu for alvb and flexible slugs for custom pages

mikkelklokkerud:
As requested by Chris:

- Menu becomes dynamically generated from sanity input (Settings view)
- Custom pages are no longer locked under a subpage category like "/products/..."
- Removes locale buttons from appearing on mobile


klyngen (approved):
I approve, but I have not spun up the code. I trust that this has been well tested. 

Usually for code that is "critical", like the navigation there should be unit-tests for testing different inputs of rows and columns.

PR #824, opened 2022-12-12
blog meta title

mikkelklokkerud:


packages/website/src/pages/blogg.js
```diff
@@ -36,8 +38,8 @@ const Blog = ({ data, serverData }) => {
   return (
     <Layout
       whiteIcons
-      pageTitle={serverData.articles.pageMetadata.pageTitle}
-      pageDescription={serverData.articles.pageMetadata.pageDescription}
+      pageTitle={pageMetaData.pageTitle}
```
klyngen, line 41: Can these attributes be null or undefined? Maybe they should be like this. Or add required to the sanity schema:
```
      pageTitle={pageMetaData?.pageTitle}
      pageDescription={pageMetaData?.pageDescription}
```

mikkelklokkerud: I made it required from sanity

klyngen (approved):

PR #834, opened 2023-01-10
add job application email form

mikkelklokkerud:
Replaces airtable form embed for with a customised form + sendgrid solution.

I think the Github Action for Sanity deployment will fail as it's not configured to deploy changes to the schema that it sees as breaking. This will in turn cause the frontend to fail because the frontend code will expect an updated schema but it will find the old one. 

Suggestion: for the PR review environments we can allow the sanity github action to update its schema by adding "--dataset development --force" to the "sanity graphql deploy" command. For merge into master we can add only the "--force". Or maybe there are other ways around this

### DIFF
--- a/packages/cms/sanity.json
+++ b/packages/cms/sanity.json
@@ -5,7 +5,7 @@
   },
   "api": {
     "projectId": "mnr37rl0",
-    "dataset": "production"
+    "dataset": "development"
   },
   "plugins": [
     "@sanity/base",
--- a/packages/mail-api/src/index.js
+++ b/packages/mail-api/src/index.js
@@ -2,9 +2,11 @@ const express = require('express');
 const bodyParser = require('body-parser');
 const cors = require('cors');
 const sgMail = require('@sendgrid/mail');
+const formidable = require('formidable');
 const dotenv = require('dotenv');
 const app = express();
 const port = 80;
+const fs = require('fs');
 
 app.use(bodyParser.json());
 app.use(bodyParser.urlencoded({ extended: true }));
@@ -16,7 +18,7 @@ dotenv.config({
 
 sgMail.setApiKey(process.env.SENDGRID_API_KEY);
 
-app.get('/', (req, res) => {
+app.get('/', (_, res) => {
   res.send('Hello');
 });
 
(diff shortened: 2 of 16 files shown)


klyngen:
💻 https://temp-834.review.test-alv.no

klyngen:
💻 https://temp-834.review.test-alv.no

packages/cms/sanity.json
```diff
@@ -5,7 +5,7 @@
   },
   "api": {
     "projectId": "mnr37rl0",
-    "dataset": "production"
+    "dataset": "development"
```
mikkelklokkerud, line 8: To make sure we don't make unwanted changes to production content

klyngen:
💻 https://temp-834.review.test-alv.no

klyngen:
💻 https://temp-834.review.test-alv.no

klyngen:
💻 https://temp-834.review.test-alv.no

klyngen:
💻 https://temp-834.review.test-alv.no

klyngen:
💻 https://temp-834.review.test-alv.no

packages/mail-api/src/index.js
```diff
@@ -58,6 +61,104 @@ app.post('/send', (req, res) => {
       console.error(error);
       res.sendStatus(500);
     });
+
+  res.sendStatus(200);
+});
+
+// job application forms
+app.post('/jobApplication/send', (req, res) => {
+  let errorMsg = '';
```
klyngen, line 70: Is it possible to split this code into smaller pieces of code. Some suggestions:
-  would be to have one function responsible of building the email message
- One function for validating the attatchments. ```
  const error = validateEmailAttatchment(file);
  
  if (error !== null)  {
    return error;
  }
  
  attatchments.push.....
```
- maybe move the new functions into a new file containing utils

File: packages/website/src/components/applyForm/applyForm.spec.js [hunk 3]
```diff
@@ -0,0 +1,26 @@
+import '@testing-library/jest-dom/extend-expect';
+import { fireEvent, render, screen } from '@testing-library/react';
+import React from 'react';
+import ApplyForm from '.';
+
+describe('applyForm-component', () => {
+  render(<ApplyForm jobTitle="Utvikler" />);
+
+  const nameInputEl = screen.getByTestId('name-input');
+  const emailInputEl = screen.getByTestId('email-input');
```
klyngen, line 10: Tester er nice :+1:

packages/website/package.json
```diff
@@ -38,6 +38,7 @@
     "react-dom": "^17.0.1",
     "react-headroom": "^3.1.0",
     "react-helmet": "^6.1.0",
+    "react-icons": "^4.7.1",
```
klyngen, line 41: I see that we are using two icons from this package. Instead of adding more packages, could we add the icons to the icon-component?

packages/mail-api/utils.js
```diff
@@ -0,0 +1,14 @@
+const allowedDocTypes = [
+  {
+    mimetype: 'application/pdf',
```
klyngen, line 3: Should this file be located under `src`?

klyngen:
Jeg får internal server error fra denne URL'en https://temp-834.review.test-alv.no/jobbe-i-alv/

packages/mail-api/src/index.js
mikkelklokkerud: I agree

klyngen:
💻 https://temp-834.review.test-alv.no

klyngen:
💻 https://temp-834.review.test-alv.no

klyngen (approved):
Looking good now :+1:

PR #842, opened 2023-01-26
add field for choosing career page image

mikkelklokkerud:


packages/cms/schemas/careerPage.js
```diff
@@ -20,6 +20,14 @@ export default {
         maxLength: 160,
       },
     },
+    {
+      name: 'mainImage',
```
klyngen, line 24: Should this be required? Does not look like the javascript-code expects that this value can be `null` or `undefined`

mikkelklokkerud: made it required 👍

klyngen:
💻 https://fix-career-842.review.test-alv.no

klyngen (approved):
Looks good :+1:

PR #845, opened 2023-01-30
Refactor richtext-types

mikkelklokkerud:
Refactor richtext component and its serialisers


klyngen:
💻 https://video-cta-845.review.test-alv.no

packages/shared-components/src/components/navItems/index.js
```diff
@@ -47,7 +47,7 @@ export const LargeTitleDropdown = ({ title, mobileDropdown, children }) => {
           className={styles.subTitleInput}
           onChange={toggleDropdown}
         />
-        <li className="list-none eight:block w-full">
+        <li className="list-none eight:block w-full flex justify-between">
```
klyngen, line 50: All `li`-tags should be surrounded by a `ul`-tag. This is not guaranteed to work in all browsers

packages/website/src/config.js
```diff
@@ -10,7 +10,7 @@ switch (config.NODE_ENV) {
       SANITY_PROJECT_ID: 'mnr37rl0',
       SANITY_DATASET:
         config.SANITY_DATASET || process.env.SANITY_DATASET || 'production',
-      SANITY_TAG: 'default',
+      SANITY_TAG: process.env.SANITY_TAG || 'default',
```
klyngen, line 13: This can probably be integrated in the review-pipeline to simplify development. It can be as simple as using the issue-number or similar.

klyngen:
Fix the semantic. This Pull Request is quite a lot larger than it should be. Personally I would split the serilization and shema changes to make it easier to roll out the small changes.  

It is nice to see technical debt disappear :+1:

packages/shared-components/src/components/navItems/index.js
mikkelklokkerud: Old fun 👏 Semantically it's even wrong to use a li tag here 🙈 since the <li> element doesn't have any sibling <li> elements..

klyngen:
💻 https://video-cta-845.review.test-alv.no

PR #846, opened 2023-01-31
introduce brands section

mikkelklokkerud:
Create brands section as per figma design and source data from new sanity fields.


klyngen:
💻 https://brandsw-846.review.test-alv.no

klyngen:
💻 https://brandsw-846.review.test-alv.no

packages/shared-components/src/index.js
```diff
@@ -1,39 +1,38 @@
-export * as Form from './components/form';
```
TheKnarf, line 1: A few to many changes in this file perhaps that are unrelated to the PR?

mikkelklokkerud: Yes, my linter decided to throw a party

mikkelklokkerud: It decided to organise imports alphabetically - this change is now undone

klyngen:
💻 https://brandsw-846.review.test-alv.no

packages/website/src/pages/om-oss.js
```diff
@@ -68,6 +75,25 @@ async function getAboutContent() {
         aboutPage: allAboutPage {
           pageDescription
           pageTitle
+          brands {
```
klyngen, line 78: The query for brands should be extracted to a constant. That will make it easier to maintain in the future.  Basically like this:

```typescript
const brandQuery = `brands {
            headingRaw
            buttonText
            buttonLink
            logos {
              alt
              image {
                asset {
                  id: _id
                  metadata {
                    dimensions {
                      height
                      width
                    }
                  }
                }
              }
            }
          }`;

```

const someQuery = gql`
  ...otherGqlStuff
  ${brandQuery}
`

packages/cms/schemas/brands.js
```diff
@@ -0,0 +1,29 @@
...
+      of: [{ type: 'imageWithAlt' }],
+      validation: (Rule) => Rule.required().min(1),
+    },
+    {
+      title: 'Button text',
+      name: 'buttonText',
+      type: 'string',
+    },
+    {
+      title: 'Button link',
```
klyngen, line 24: Does it make sense to have a link without a URL? Should this be required?

packages/website/src/pages/om-oss.js
mikkelklokkerud: Is this a new pattern we want to adopt? Or are we doing this already somewhere in our codebase - if so it would be useful to look at where/how we store and possibly import the query contants. Can think of a few various ways

mikkelklokkerud: Regardless, it's a good idea.. I want a codemod to apply this everywhere in the codebase....

klyngen: No, don't think it is a pattern. But put them somewhere central, maybe even close to the hookQueries. You can call them queryFragments or something. That will greatly improve the codebase :+1:

klyngen:
💻 https://brandsw-846.review.test-alv.no

klyngen (approved):

klyngen:
💻 https://brandsw-846.review.test-alv.no

PR #850, opened 2023-02-03
Add linkable headings to service pages

mikkelklokkerud:
Update styling to linkable headings and add it to service pages


klyngen:
As a note to myself. I see that we don't deploy any review-environment in the case of the shared-library being modified :thinking:

klyngen (approved):

PR #854, opened 2023-02-08
fix employees page mobile bug

mikkelklokkerud:
Employee cards weren't showing on initial page load before interactivity on mobile. Now they do

### DIFF
--- a/packages/shared-components/src/components/employeeGroup/index.js
+++ b/packages/shared-components/src/components/employeeGroup/index.js
@@ -34,14 +34,14 @@ export const EmployeeGroup = ({
   };
   // Close employee bio section by clearing activeBio state
   const handleCloseClick = () => {
-    // eslint-disable-next-line no-self-assign
     updateActiveBio(null);
   };
+
   return (
     <>
       <div>
-        <div class="md:flex justify-center">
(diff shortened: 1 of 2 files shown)


packages/shared-components/src/components/employeeGroup/index.js
```diff
@@ -34,14 +34,14 @@ export const EmployeeGroup = ({
...
-    // eslint-disable-next-line no-self-assign
     updateActiveBio(null);
   };
+
   return (
     <>
       <div>
-        <div class="md:flex justify-center">
-          <div className="grid xs:gap-4 justify-center gap-y-4 five:gap-y-0 xs:mx-auto mb-4 xs:max-w-570 seven:max-w-none nine:max-w-grid -mx-4 xs:mx-0 grid-cols-employees-xs xs:grid-cols-employees-sm seven:grid-cols-employees-md nine:grid-cols-employees-lg">
+        <div class="md:flex justify-center ">
```
klyngen, line 43: Just a tip. If you set up the correct language-servers on the page these kind of spaces gets removed automatically. I am a bit uncertain what that plugin is called in vscode, but if I have it in emacs, you can have it too :wink:

klyngen (approved):
Looks good with only a tip as a comment

PR #856, opened 2023-02-08
Fixed 502 when loading page

### LINKED ISSUE #855: 502 responses after startup is being cached
Do some discovery into how we can stop caching 502 responses coming from nginx. Maybe even make the 502 site a little nicer. Maybe a page with a 10 second retry would be a good solution.

klyngen:
Varnish has made deploying the webpage a nightmare
- Made sure varnish does not cache 500-pages
- Made a simple custom 500-page that retries while the user waits for the actual page to load
- Cleaned up the repo a bit and moved nginx-config-files into its own folder

closes #855


mikkelklokkerud (approved):
Najs med 502 fix og custom 500 page 🪄

PR #859, opened 2023-02-13
add support for file uploads with clamav scanning

mikkelklokkerud:
Tested with clamav scan solution set up by Anders


klyngen:
💻 https://scan-file--859.review.test-alv.no

packages/mail-api/src/index.js
```diff
@@ -99,34 +106,74 @@ app.post('/jobApplication/send', (req, res) => {
...
           return;
         }
 
-        // Upon successful validation - convert file to base64 and add to attachments array
-        const base64content = await readFile(filepath, { encoding: 'base64' });
+        // scan file with clamav
+        const formData = new FormData();
+        formData.append('testfile', createReadStream(files[file].filepath));
+
+        const virusScanResponse = await fetch(
```
klyngen, line 136: Litt samme greia som sist. Ta gjerne antivirus-koden inn i en egen metode ala `scanFileForMaliciousContent()` som gjerne bare returnerer `true` eller `false`. På den måten blir koden myyyye lettere å håndtere.

packages/mail-api/src/index.js
```diff
@@ -99,34 +106,74 @@ app.post('/jobApplication/send', (req, res) => {
...
+        // If virus is detected, send email with details to itadmin
+        if (virusScanResponse[0] === 'FOUND') {
+          msg = generateVirusEmail(virusScanResponse[1], email);
+          virusDetected = true;
+
+          return;
+        }
+
+        // log to server console
+        console.log('No virus detected.');
```
klyngen, line 156: This console message should maybe have some more parameters to be useful. Maybe the senders email or the name of the attatchment. If not, this is not very useful.

I usually add correlationId when I make decent logs

packages/mail-api/src/index.js
```diff
@@ -99,34 +106,74 @@ app.post('/jobApplication/send', (req, res) => {
...
           disposition: 'attachment',
         });
+
+        // Add attachments to mail object
+        msg.attachments = attachments;
       }
     }
 
     if (errorMsg) {
+      console.error(errorMsg);
```
klyngen, line 176: Here also, try to give the logging some context to make things easier to debug on a rainy day

packages/mail-api/src/index.js
```diff
@@ -99,34 +106,74 @@ app.post('/jobApplication/send', (req, res) => {
...
       const attachments = [];
 
       // Convert files to array if only one file is uploaded
       [files].length === 1 && (files = [files]);
 
       for (const file in files) {
-        const { filepath } = files[file];
-        const attachmentErrors = validateEmailAttachment(files[file]);
+        // If virus is detected, stop processing
+        if (virusDetected) {
```
klyngen, line 120: Don't think this if-block is needed. If you find a virus, you return on line 150. Kill this block

klyngen: You can probably also kill the virusDetected-variable, if I read the flow of the code correctly :thinking:

packages/mail-api/src/utils.js
```diff
@@ -42,4 +42,14 @@ const validateEmailAttachment = async (file) => {
   return errors;
 };
 
-module.exports = { validateEmailAttachment };
+const fetch = (url, options) =>
+  import('node-fetch').then(({ default: fetch }) => fetch(url, options));
+
+const generateVirusEmail = (threat, sender) => ({
```
klyngen, line 48: Nice to put this in it's own method. Makes the code easy to read :+1:

klyngen (commented):
Hey! La igjen noen kommentarer på PR'en. Litt sånn småtteri. Ellers bra greier :+1:

klyngen:
Hey! Denne branchen ser litt glemt ut. Denne er vel klar til å merges etter at du har gjort noen små justeringer på koden :smile:

klyngen:
💻 https://scan-file--859.review.test-alv.no

packages/mail-api/src/index.js
mikkelklokkerud: The purpose behind this if-block is to prevent unnecessary processing of attachment files in cases where one of the previously scanned attachment files turned out to be malicious.

In a case where we find out that one file is malicious and another one is not, I would say it's better to assume none of the attachment files are safe anyway.

I improved the comment above line 119 to make that purpose clearer

Thoughts?

klyngen:
💻 https://scan-file--859.review.test-alv.no

klyngen:
💻 https://scan-file--859.review.test-alv.no

klyngen:
💻 https://scan-file--859.review.test-alv.no

packages/mail-api/src/index.js
klyngen: You have a return on line 143 that prevents that loop from running again. That if-statement is actually just a waste of computation.

mikkelklokkerud: wait, I take it back and I agree with you on this 😂 I thought return would close the current iteration but not stop the loop, like how `continue` works.. 

That said, in addition to stopping the loop, `return` will of course also terminate the method it's called within, which is not a good idea here. 

I'll update to use `break` instead of `return` (143) and remove the if statement /w variable on top. That will give the effect I wanted here.

mikkelklokkerud: and make it more readable. I never liked that variable on top :D

klyngen:
💻 https://scan-file--859.review.test-alv.no

klyngen:
💻 https://scan-file--859.review.test-alv.no

klyngen:
💻 https://scan-file--859.review.test-alv.no

PR #861, opened 2023-02-20
improve prop names and fix blogg title color bug

mikkelklokkerud:
Fix blog header color bug and improve props naming for blog carousel


klyngen:
💻 https://blogg-titl-861.review.test-alv.no

PR #869, opened 2023-02-27
mail-cta-update

mikkelklokkerud:


packages/shared-components/src/utils.js
```diff
@@ -52,20 +52,23 @@ export const createFormData = (inputValues) => {
...
 export const submitWithDelay = async (apiUrl, body) => {
-  fetch(apiUrl, {
+  const delay = new Promise((resolve) => setTimeout(resolve, 400));
+
+  let submissionResponseStatus;
+  const trySubmit = fetch(apiUrl, {
     method: "POST",
     body,
   })
     .then(() => {
```
klyngen, line 62: I believe the success / error conditions needs a bit more code to work properly. 
```javascript
fetch(...something)
.then(response => {
  // This can be 200 and 400 response
  
  if(!response.ok) {
    return "error";
  }
  return "success";
  
}).catch(response => {
  // usually CORS
  // sometimes 500
});

```

mikkelklokkerud: made an update 👍

klyngen (approved):
Looks perfect :+1:

PR #879, opened 2023-03-30
Career page tweaks

klyngen:
# Career page tweaks
- VideoList was crashing the page. Put check to see if we were on an IOS device into a useEffect that only runs once.
- Font-weight tweaks
- Adjusted order of image + icon

### DIFF
--- a/packages/website/src/components/videoList/index.js
+++ b/packages/website/src/components/videoList/index.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
 import { BlockContent } from "shared-components";
 
 const VideoList = ({ title, description, list, config }) => {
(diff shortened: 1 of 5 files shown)


packages/website/package.json
```diff
@@ -19,7 +19,7 @@
     "browser-monads": "^1.0.0",
     "classnames": "^2.2.6",
     "crypto": "^1.0.1",
-    "fastify": "4.9.2",
+    "fastify": "latest",
```
TheKnarf, line 22: Better if we refer to the actual version number.

packages/website/src/components/videoList/index.js
```diff
@@ -28,7 +28,12 @@ const VideoList = ({ title, description, list, config }) => {
 };
 
 const Video = ({ videoWebM, videoMp4 }) => {
-  const isIOS = navigator.userAgent.match(/(iPhone|iPad)/);
```
TheKnarf, line 31: Couldn't this have been:

const isIOS = typeof window !== "undefined" ? navigator.userAgent.match(/(iPhone|iPad)/) : "";

mikkelklokkerud (approved):

packages/website/package.json
klyngen: Removed change

TheKnarf (approved):

PR #880, opened 2023-04-01
return not found page if no serverData or if page is draft

mikkelklokkerud:


klyngen (changes requested):
I think this solution works, but only for `jobbe-i-alv`. 
Put the same component into a `page/500.ts`-file. See the documentation below
https://www.gatsbyjs.com/docs/how-to/adding-common-features/add-500-page/

PR #896, opened 2023-05-19
Fix: tracking forms using plausible

klyngen:
This changes how we handle events when using forms
- Plausible wraps event-listeners on forms when applying custom events to it. The code on forms are known to be buggy.
- transitioned to fire events manually instead using `trackCustomEvent`-method


TheKnarf (approved):

klyngen:
💻 https://tracking-f-896.review.test-alv.no

PR #901, opened 2023-06-15
No longer showing unpublished job-listings

klyngen:
We now show a 500 page whenever someone hits an unpublished page
- Transitioned `jobbe-i-alv`-pages to use GROQ instead of graphql due to limitations from the sanity API. Benefit is that the API is much easier to use and queries are very easy to read
- Added a 500 page
- Modified gatsbyImage-integration to work with image-data retrieved by GROQ queries
- Added jsdoc-typings to configuration
- Added a simple sanityClient to make it easier to use in the future

### DIFF
--- a/packages/website/package.json
+++ b/packages/website/package.json
@@ -14,6 +14,7 @@
   "dependencies": {
     "@apollo/client": "^3.5.8",
     "@sanity/block-content-to-react": "^2.0.7",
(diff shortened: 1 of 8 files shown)


klyngen:
💻 https://unpublishe-901.review.test-alv.no

klyngen:
💻 https://unpublishe-901.review.test-alv.no

PR #963, opened 2023-09-04
962: add max file size to form parser

Fordiboii:


klyngen:
💻 https://962-extend-963.review.test-alv.no

egilsebastiandahl (approved):
Ser bra ut for min del!

PR #971, opened 2023-09-22
Service page touches 📦

klyngen:
Small changes to fonts
- Service page title is smaller and suits smaller devices better
- Hyphens removed from service page liks component
- Image flip has higher font weight


yvindo (approved):

klyngen:
💻 https://service-pa-971.review.test-alv.no

PR #976, opened 2023-10-19
Feature #975 🍪

### LINKED ISSUE #975: Add google tag manager and linkedin cookies
We want to be able to see conversions to our website from google results

## What needs to be done
- Implement a cookie banner that is GDPR compliant
- Cookie mechanism that ensures that we only do tracking after the user has consented to the cookie banner

klyngen:
# Cookie banner and tracking
Due to there being very few good tracking libraries for gatsby I implemented my own GDPR compliant mechanism to handle loading of trackers.
- Supports linkedIn and google tag manager
- Only loads tracker after a user has done a consent
- Simple and responsive design

closes #975

### DIFF
--- a/packages/website/gatsby-config.js
+++ b/packages/website/gatsby-config.js
@@ -27,6 +27,13 @@ module.exports = {
   },
   plugins: [
     {
+      resolve: "alv-gdpr-tracking",
+      options: {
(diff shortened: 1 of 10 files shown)


JohanAOstbye (approved):
Looks good m8

